PR #36, opened 2018-06-15
ADD: Robust OpenDSS Parser

### LINKED ISSUE #13: Improved OpenDSS Parser
Make the parser more robust and add more comprehensive support for the components that this package supports.

### LINKED ISSUE #34: Voltage Source as Generators
Add parser options adding generators for voltage sources.  The most common example is a feeder with one voltage source at the root node and no explicit generators.

### LINKED ISSUE #35: Generator Costs
Make these apply to the sum of active powers on a multiphase generator, rather than separate functions for each phase.

With PowerModels.jl [#308](https://github.com/lanl-ansi/PowerModels.jl/pull/308) cost functions are now the same for each phase. Refactor to reflect this change.

- [x] OpenDSS Parser
- [x] Matlab Format

pseudocubic:
This represents a significant rewrite and update to portions of the OpenDSS parser. In particular, I have completely rewritten the way default properties of OpenDSS components are calculated. This is due to the interconnectedness of many of the properties, where it is possible to define the same property of a component, _e.g._ a `rmatrix`, via many different combinations of input properties. This rewrite is an effort to rally around a standardized set of properties that ThreePhasePowerModels can assume exists during the creation of new components. This has been achieved by creating constructor functions for each supported component.

In order to more easily develop and debug this code, I have refactored `src/io/opendss.jl` into three separate files, with `opendss.jl` still containing the conversion from raw parsed `dss_data` to `tppm_data` structured for use by ThreePhasePowerModels, `dss_parse.jl` containing all functions that perform specific parsing tasks of raw data, and `dss_structs.jl` containing all constructor functions for the different supported OpenDSS components.

I have also made the actual parsing more robust, including the support of Reverse Polish Notation math within property values. All array delimiters are now supported which include `" ' () [] {}`.

All property names for every OpenDSS component, in order, have been added, so that any component is able to at least be parsed correctly into the raw `dss_data` dictionary via the `parse_dss` function.

The `sourcebus`, which is created by the `new circuit` command, required in valid OpenDSS, instances a generator, whose stiffness can be controlled using VSource properties, and a bus.

This PR also adds more transformer support, but this is still a work in progress. Three winding transformers are currently added by creating a `starbus` and connecting each winding to that starbus. Transformer values are not populated yet.

The OpenDSS parser has been validated again OpenDSS using two 3-bus cases, balanced and unbalanced.

In future (not in this PR) more unit tests to cover the many different cases covered by the constructor functions, but for now these features should be considered developmental. Preferred method for many components should be to specify the matrices, _e.g._ `cmatrix`, `xmatrix`, `rmatrix` in Lines and Linecodes. 

Closes #13 
Closes #34 
Closes #35


pseudocubic:
@bluejuniper when you have some time, I could use input on formulating the values for the `"sourcebus"` generator that I mention above. This is related to #34.

The function we are trying to complete is in `opendss.jl`, lines 243-301, and the values available to fill the necessary fields can be found in `dss_structs.jl`, lines 655-692

pseudocubic:
As a note on the progress. I now have a generator to represent `sourcebus`, and I fixed an issue with branch impedances not being in per unit, as PowerModels expects.

3-bus balanced and unbalanced cases were added to the test suite, but currently only checks that OPF is solvable in each of AC Polar, SOC, and Linear DistFlow. Validating against OpenDSS will still take some work.

pseudocubic:
I'm not sure why the test is failing. It passes on my personal travis account: https://travis-ci.com/pseudocubic/ThreePhasePowerModels.jl/builds/76872189

src/io/common.jl
```diff
@@ -21,6 +21,15 @@ function parse_file(file::String; import_all::Bool=false, vmin::Float64=0.9, vma
 end
 
 
+""
+function parse_file(filename::String; kwargs...)
+    tppm_data = open(filename) do io
+        parse_file(io; kwargs...)
```
ccoffrin: should this be `parse_file`?  Seems more general, like `parse`.

src/io/opendss.jl
```diff
@@ -862,23 +146,22 @@ function dss2tppm_load!(tppm_data::Dict, dss_data::Dict, import_all::Bool)
...
             nphases = tppm_data["phases"]
             name, nodes = parse_busname(defaults["bus1"])
 
-            loadDict["name"] = defaults["id"]
+            loadDict["name"] = defaults["name"]
             loadDict["load_bus"] = find_bus(name, tppm_data)
-            loadDict["pd"] = PMs.MultiPhaseVector(phase_on_off(fill(defaults["kw"], nphases), nodes, nphases))
-            loadDict["qd"] = PMs.MultiPhaseVector(phase_on_off(fill(defaults["kvar"], nphases), nodes, nphases))
+            loadDict["pd"] = PMs.MultiPhaseVector(parse_array(defaults["kw"] / 1e3, nodes, nphases))
+            loadDict["qd"] = PMs.MultiPhaseVector(parse_array(defaults["kvar"] / 1e3, nodes, nphases))
```
ccoffrin: Where does this `1e3` some from?  To convert `kw` into `mw`?

pseudocubic: yes

src/io/common.jl
pseudocubic: well, I would say that because it only expects file paths, `parse_file` would be the more appropriate name for this one, and the one that takes `IOStream`s would then become `parse` if you wish to follow that kind of convention.

ccoffrin: Yes, I meant to refer to the function called in line 27, `parse_file(io; kwargs...)` which takes an io as input.

src/io/opendss.jl
pseudocubic: I haven't tried that yet, but since the minimal set of components so far are already converted, I would propose keeping it as is, and I can make a new Issue about exploring this possibility.

src/io/common.jl
pseudocubic: I just tried to do this, and it somehow completely breaks the internal `parse(Type, Str)` command which most of the code relies upon to parse things, so I propose we leave it as is for now, and I can see if I can get it to work later.

src/io/opendss.jl
ccoffrin: Sounds good to me.

src/io/common.jl
ccoffrin: Fine by me.

PR #50, opened 2018-07-05
FIX: OpenDSS Parser, PowerFlow Problem

pseudocubic:
Fixes parsing of OpenDSS lines. Units of `rmatrix`, `xmatrix`, `cmatrix`,
`b_fr` and `b_to` were 1/MVA, instead of pu, fixed now. Adds factor 3
to `rmatrix` and `xmatrix`. Refactors handling of `Rg`, `Xg` in
`createLine`.

Fixes bug in bus parsing; vmin and vmax were being overwritten in the
for-loop, changed variable names to fix this.

Adds unit tests to ensure better parity with results of OpenDSS. Adds
new 2-bus test with only diagonal elements in `rmatrix` and `xmatrix`.
Tightens tolerances on existing unit test on 3-bus balanced and
unbalanced cases. Adjusts simple test case to move generator from
sourcebus to bus 2.

Fixes a typo in `src/prob/tp_pf.jl` that was causing PowerFlow problems
based on OpenDSS data to fail.

Adds new powerflow tests based on OpenDSS data.

### DIFF
--- a/src/form/acp.jl
+++ b/src/form/acp.jl
@@ -27,16 +27,16 @@ function constraint_ohms_tp_yt_from(pm::GenericPowerModel{T}, n::Int, h::Int, f_
     va_fr = [var(pm, n, j, :va, f_bus) for j in PMs.phase_ids(pm)]
     va_to = [var(pm, n, j, :va, t_bus) for j in PMs.phase_ids(pm)]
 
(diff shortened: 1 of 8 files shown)


ccoffrin:
These changes look fine to me.  @frederikgeth, make sure to build your PR off this base.

PR #52, opened 2018-07-10
Change of phase order at reference bus

frederikgeth:
This PR is not yet ready, unit tests have only been updated partially.

Some feedback is welcome on the following before I finish:
- where to put the helper functions `wrapto180` and `wraptopi`?, which limit angles to [-180, 180] and [-pi, pi] respectively.
- now the OpenDSS parser sets the voltage angle reference in the order 0, -120, -240 degrees, which is then wrapped using the function above.

What to do with the buses with fixed voltage magnitude? In the matpower-style input format it isn't possible to define such a bus. I'm not in favor of using the voltage magnitude bounds to enforce this, as it has a different meaning. E.g., even if a voltage source is fixed at a certain output voltage, that value should still be within the nodal voltage bounds. 

How do you want to maintain compatibility with the matpower style inputs?


src/core/constraint_template.jl
```diff
@@ -87,7 +87,17 @@ end
...
 function constraint_tp_theta_ref(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw, ph::Int=pm.cph)
-    constraint_tp_theta_ref(pm, nw, ph, i)
+    bus = ref(pm, nw, :bus, i)
+    varef = bus["va"][ph]
+    constraint_tp_theta_ref(pm, nw, ph, i, varef)
+end
+
+
+""
+function constraint_tp_voltage_magnitude_ref(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw, ph::Int=pm.cph)
```
ccoffrin: Is this different than `constraint_voltage_magnitude_setpoint`?

src/core/constraint_template.jl
```diff
@@ -87,7 +87,17 @@ end
 
 ""
 function constraint_tp_theta_ref(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw, ph::Int=pm.cph)
-    constraint_tp_theta_ref(pm, nw, ph, i)
+    bus = ref(pm, nw, :bus, i)
+    varef = bus["va"][ph]
+    constraint_tp_theta_ref(pm, nw, ph, i, varef)
```
ccoffrin: To match the semantics of the refrence bus in transmission systems, this constraint should fix the phase angles to standardized values that only depend on phase/conductor configuration.  If you want to fix the angles to specific values in the data, then the constraint should have a different name.  However, I am curious in what use case such a constraint is useful.

src/prob/tp_opf.jl
```diff
@@ -33,6 +33,7 @@ function post_tp_opf(pm::GenericPowerModel)
 
         for i in ids(pm, :ref_buses)
             constraint_tp_theta_ref(pm, i, ph=h)
+            constraint_tp_voltage_magnitude_ref(pm, i, ph=h)
```
ccoffrin: This looks like power flow semantics to me rather than optimal power flow semantics.  I know is is typical to fix the root of a feeder to 1.0 V p.u., but isn't it possible to vary that set point using the step-down transformer that is feeding it?

src/prob/tp_opf_bf.jl
```diff
@@ -32,6 +32,7 @@ function post_tp_opf_bf(pm::GenericPowerModel)
     for h in PMs.phase_ids(pm)
         for i in ids(pm, :ref_buses)
             constraint_tp_theta_ref(pm, i, ph=h)
+            constraint_tp_voltage_magnitude_ref(pm, i, ph=h)
```
ccoffrin: Same remark as above.

src/core/multiphase.jl
```diff
@@ -0,0 +1,7 @@
+function wrapto180(degrees)
+    return degrees - 360*floor((degrees + 180)/360)
+end
+
+function wraptopi(radians)
+    return radians - 2*pi*floor((radians+pi)/(2*pi))
+end
```
ccoffrin: Unless they are used elsewhere, I would put these in `io/opendss` or `io/common` for now.

ccoffrin:
Regarding your remark,

> matpower-style input format it isn't possible to define such a bus.

I think the confusion is on the semantics of OPF vs PF.  My inclination is to keep the basic OPF very similar to the transmission system one.  If you think OPF variants are needed for distribution systems then you can define new problem specifications that are a mix of the OPF and PF constraints, as you see fit.

Note that bus type 3 in the matlab format could be used for specifying a refrence bus with a fixed voltage angle and magnitude.

src/core/constraint_template.jl
frederikgeth: For a three-phase case an explicit neutral, a voltage reference vector abcn may look like [Va, Vb, Vc, 0] (similar to (6) in [1]). So four conductors but phases 120 degrees shifted. In split-phase the angles may be 90/180 degrees. I'm not sure what to do to keep compatibility, except by predefining a set of possible configurations (e.g. 3-phase no neutral, 3-phase explicit neutral, split-phase, etc.) and how they map to a reference bus definition?  

[1] Ciric, R. M., Feltrin, A. P., & Ochoa, L. F. (2003). Power flow in four-wire distribution networks - general approach. IEEE Trans. Power Syst., 18(4), 1283–1290.

src/prob/tp_opf.jl
frederikgeth: There is a challenge related to recovering node voltage variables in the original variable space. when you only fix the angles at the reference bus. It may not be possible to recover a solution if W is not rank-1, and you need that as a starting point to run algorithm 1 or 2 in Gan and Low.

Gan and Low can be made to work with just fixing the voltage angles, if the voltage angles are 120 degrees and the voltage magnitudes are identical for the three phases. Your W matrix variable for W=U*U' with U = [um/_0, um/_120, um/_0, ] and w = um^2 will then look like
```
W = 
[ w, -w/2 -w/2;
-w/2, w, -w/2;
-w/2, -w/2, w] 
+ j * sqrt(3)/2*
[0, w, -w;
-w, 0, w;
w, -w, 0]
```

We'll have to work out other reference bus cases where the angles are not evenly spaced / magnitudes are not identical, but rank-1-ness may not easily be guaranteed.

src/core/multiphase.jl
frederikgeth: ok

src/core/constraint_template.jl
ccoffrin: Fixing the values from the datafile is a resonable solution while you are exploring the formulations.  In the long term me feeling is that we should add some more metadata to the network data, which indicates different conductor configurations.  This info can be used by the refrence bus constraint to set the variables correctly.

frederikgeth: Ok, good approach. So which of the two do you want now?
`@constraint(pm.model, va == wraptopi(2 * pi / nphases * (1-h)))`
or 
`@constraint(pm.model, va == varef)` with varef set by the parser?

src/prob/tp_opf.jl
ccoffrin: The issue of problem formulation and feasible solution recovery from a relaxation are orthogonal.  The standard OPF problem provides no guarantees of feasible solution recovery.  If you would like to propose an OPF variant, which has such properties, that is fine just provide a doc string explaining the motivation for the formulation and a citation if it makes sense.

src/core/constraint_template.jl
ccoffrin: My recommendation would be to,
1) `constraint_tp_theta_ref` to follow the angle offset standard you think is best.  I believe that is, `@constraint(pm.model, va == wraptopi(2 * pi / nphases * (1-h)))`
2) add a new constraint (e.g. `constraint_mc_theta_setpoint`) for multi-conductor theta setpoint, which uses `@constraint(pm.model, va == varef)` as specified in the data file.
3) Add a new problem specification (e.g. `mc_opf` or `tpn_opf`), which supports an explicit neutral and uses `constraint_mc_theta_setpoint` instead of `constraint_tp_theta_ref`

frederikgeth:
I had Gan and Low's unbalanced OPF problem definition in mind (where a voltage magnitude reference is included). I will remove the voltage magnitude aspects. 

Nevertheless, down the line it may be useful to support such bus types for the OPF problem type or define a new OPF problem type with support for fixed voltage angle + magnitude buses. As far as I've seen published OPF results by Gan&Low and Dall'Anese set both the angle and the voltage magnitude at the source bus.

frederikgeth:
Did the minimum fixes for now. I can't run `runtests.jl` without errors though. But I don't think I broke the opendss and matlab parsers though.

frederikgeth:
The power flow results based on the NLP formulation seems consistent with OpenDSS based on very limited tests, but I'll have to do a bit more exploration. I'm pretty confident though that the linearised (simplified) branch flow model isn't consistent with any of the published models [1] or [2].

[1] Sankur, M. D., Dobbe, R., Stewart, E., Callaway, D. S., & Arnold, D. B. (2016). A linearized power flow model for optimization in unbalanced distribution systems, (June).
[2] Gan, L., & Low, S. H. (2014). Convex relaxations and linear approximation for optimal power flow in multiphase radial networks. In PSSC (pp. 1–9). Wroclaw, Poland. https://doi.org/10.1109/PSCC.2014.7038399

ccoffrin:
Merging this, @pseudocubic has fixes for the tests.

PR #57, opened 2018-07-16
Debugging of linearised branch flow model

frederikgeth:
I'm trying to validate the simplified/linearised branch flow model to match [1]. Wondering what you think of this.

TODOs:
- [x] re-derive `constraint_tp_voltage_magnitude_difference` to take into account the changed phase order.
- [x] support in-line transformers or not?
- [x] update unit tests

[1] Sankur, M. D., Dobbe, R., Stewart, E., Callaway, D. S., & Arnold, D. B. (2016). A linearized power flow model for optimization in unbalanced distribution systems


frederikgeth:
If you consider adding this PR we may need to think of a formulation name that also refers to the fact it represents a linearisation of the unbalanced BFM formulation (as in PMs `SOCBFPowerModel` it is the relaxed formulation, not the linearised one).

src/form/bf.jl
```diff
@@ -31,9 +35,28 @@ function constraint_tp_voltage_magnitude_difference(pm::GenericPowerModel{T}, n:
...
+function constraint_tp_flow_losses(pm::GenericPowerModel{T}, n::Int, c::Int, i, f_bus, t_bus, f_idx, t_idx, r, x, g_sh_fr, g_sh_to, b_sh_fr, b_sh_to, tm) where T <: PMs.AbstractBFForm
+    p_fr = var(pm, n, c, :p, f_idx)
+    q_fr = var(pm, n, c, :q, f_idx)
+    p_to = var(pm, n, c, :p, t_idx)
+    q_to = var(pm, n, c, :q, t_idx)
+    w_fr = var(pm, n, c, :w, f_bus)
+    w_to = var(pm, n, c, :w, t_bus)
+
+    @constraint(pm.model, p_fr + p_to ==  g_sh_fr*(w_fr/tm^2) +  g_sh_to*w_to)
+    @constraint(pm.model, q_fr + q_to == -b_sh_fr*(w_fr/tm^2) + -b_sh_to*w_to)
```
ccoffrin: Is this what they proposed in the original paper?  Seems like it would quickly lead to infeasiblities to me.  Using `>=` instead of `==` would be more reasonable, but I am not inclined to depart from the original paper if we can avoid it.

frederikgeth: It is a linearised model, and follows the same principles as the balanced derivation:
- start from the relaxed formulation in the lifted variable space
- assume the current through the series element in the pi section is approximately zero 
- substitute this current variable out of the constraint linking the voltage magnitudes 
- substitute it out of the line power balance as well. The only remaining part of the line losses is are the shunt contributions. 

[1] drops the L_{mn} = z I_n (I_n)* term representing the (series) losses going from the exact model in equation (8) to (20) making it lossless. 

The step of keeping the shunt contributions isn't explicitly mentioned in [1], however [2] mentions he did model the line shunts like this:

> line shunt is modeled using the π model— assuming a fixed impedance load at each end of the line with the impedance being half of the line shunt 

[1] Sankur, M. D., Dobbe, R., Stewart, E., Callaway, D. S., & Arnold, D. B. (2016). A linearized power flow model for optimization in unbalanced distribution systems, (June).
[2] Gan, L., & Low, S. H. (2014). Convex relaxations and linear approximation for optimal power flow in multiphase radial networks. In PSSC (pp. 1–9). Wroclaw, Poland. https://doi.org/10.1109/PSCC.2014.7038399

ccoffrin: I agree your logic is sound and in the spirit of the original derivation.  However, if we do not have a precise mathematical derivation to work from we may accidentally dis-advantage the original method.

On option is to reach out to the original authors and ask what they prefer.

I'll leave the final determination of this formulation to you.  Just let me know when you are ready for it to be merged into master.

frederikgeth: For me this is ready. I can attempt a PR for the SOC relaxation of the SDP relaxation of BFM later this week if my agenda doesn't change too much. 

The only question remaining is whether we keep the name `SOCBFPowerModel`, as it is a linear approximation, not a relaxation. If you want to change it at this moment, feel free to propose a new name, otherwise go ahead and merge.

ccoffrin: I recommend the following guide lines for formulation names,
1) Give the formulations names that are reminiscent of the original source paper, add a doc string citation to the formulation type definition
2) indicate in the doc string if the formulation is an approximation or a relaxation 
3) Generally use subtypes of the formulations in PowerModes.  Notable exceptions would be ACP or ACR which have clear multi-phase analogs

I'll merge this now, you can make another PR if you would like to revise the type names.

frederikgeth: ok, great!

Issue #58, opened 2018-07-17
Harmonize formulation names, e.g. SOCBFPowerModel


frederikgeth:
Currently the `SOCBFPowerModel` is an implementation of a linear approximation of branch flow model, not a convex relaxation.

Per the guidelines [here](https://github.com/lanl-ansi/ThreePhasePowerModels.jl/pull/57), the name should be updated:

> - Give the formulations names that are reminiscent of the original source paper, add a doc string citation to the formulation type definition
> - Indicate in the doc string if the formulation is an approximation or a relaxation
> - Generally use subtypes of the formulations in PowerModes. Notable exceptions would be ACP or ACR which have clear multi-phase analogs

pseudocubic:
Per our discussions, suggest to add full formulation names as we have been doing, but also to add alias to accepted formulation names, e.g. LPUBFDiagModel is aliased to LinDist3FlowModel

pseudocubic:
This seems completed to me, is there additional work still to be done? @frederikgeth

ccoffrin:
If possible, I would recomend utilizing the BFA type name from PowerModels for one of the formulations.

frederikgeth:
I think this issue can be closed after addressing Carleton's point on the BFA form (i.e. LPDiag).

PR #60, opened 2018-07-18
SDP and SOC relaxation of BFM + further debugging

frederikgeth:
Working on adding more formulations in this PR, focusing on matrix-based mathematical models. The matrix-wise problem formulation is defined in `tp_opf_bf_mat.jl` (i.e. no outer loop over the conductors in `post_tp_opf_bf_mat`):
- SDP relaxation of balanced BFM `SDPUBFForm`
- SOC relaxation of SDP relaxation of balanced BFM
  - conic form `SOCConicUBFForm` 
  - nlp form `SOCNLPUBFForm`
- Simplfied (linear approximation) version of the lifted unbalanced formulation 
  - implementation similar to Gan and Low `LPfullUBFForm` with the 3x3 Hermitian matrix variable for the node voltages. 
  - implementation similar to Sankur et al. `LPdiagUBFForm` with only the diagonal of the Hermitian matrix variable for node voltages being retained. 

The SDP formulation is tested with Mosek and obtains the same result as the NLP AC formulation for `case3_unbalanced.dss`. It is important though to set certain bounds low enough (e.g. rate_a of lines, pmin, pmax, qmin, qmax of generators) for Mosek not to stall. 

Open questions:
- Does it make sense to keep all three linear approximation formulations? (scalar, diagonal matrix and full matrix variant for `constraint_tp_voltage_magnitude_difference_mat`. I'm not sure that the diagonal matrix and full matrix variants are fundamentally different though. TBD).
- Do you want to keep the NLP and conic variants of the SOC relaxation formulation? 
- So far I haven't tried to support 4x4 impedance, I prefer to do that in a later PR if that's acceptable. 

@hakanergun, feel free to inspect and comment. You have experience with these formulations as well :)

Sidenote, this includes a bugfix in SOCBFPowerModel.

TODOs:
- [x] remove debugging code
- [x] write unit tests
- [x] remove.values from MultiPhaseConductor objects
- [x] remove some more trivial constraints from the set of SOCs derived from the SDP relaxation
- [x] only fix voltage angle at the slack bus
- [x] automatically derive number of conductors
- [ ] be consistent with guidelines https://github.com/lanl-ansi/ThreePhasePowerModels.jl/issues/58
- [x] update naming conventions #61 
- [x] tight bounds on all variables


frederikgeth:
The problem types have been adapted today. Tomorrow I'll continue, but I have some observations which could be of interest to you:
- At the moment I'm a bit confused why the SOC relaxation of the SDP gives different results for NLP/Ipopt than SOC/Mosek. The latter returns the more accurate solution. 
- For MOSEK it is very important to set the bounds on the lines low enough (less than the default from the opendss parser), otherwise it exits with :stall (problem for both SDP and SOC).
- `case3_unbalanced.dss` needs to have the basemva set to 1 MVA; with 100 MVA solutions are not the same as the OpenDSS results copied here: https://github.com/lanl-ansi/ThreePhasePowerModels.jl/issues/47
- the SDP result is identical to the NLP and Opendss result for the case3_unbalanced to <1E-6 (losses and voltage) with that 1MVA base

ccoffrin:
Re observation 1, When you say "The latter returns the more accurate solution.", what is your criteria for more accurate?

Re observation 2, that sounds quite odd.  Weak constraints should not typically cause any convergence issues.  The issue may be more about problem scaling.

Re observations 3 and 4, that's great!

frederikgeth:
- The NLP/Ipopt SOC relaxation exhibits negative line losses if you drop the valid inequality `trace(p_fr) + trace(p_to) >=0`. The SOC/Mosek SOC relaxation does not have this issue, even without that valid inequality. 
- I feel it may to do with scaling (including having to adapt the basemva), but I'll have to dig deeper today.

ccoffrin:
Regarding your first point, there are two possibilities,
1) There are multiple optimal solutions to the mathematical optimization problem.  In this case, each solver will pick one optimal solution more or less at random.  This can be confirmed by showing that both solvers objective values are the same (up to some tolerance); and by adding additional terms to the objective function to set a preference for some optimal solutions over others.
2) The optimal solution to the mathematical optimization problem is unique.  In this case, the discrepancy can only be due to numerical issues and the solver parameters should be adjusted to make the results more comparable.

For what it is worth, my experience thus far has been that Ipopt is the most accurate and robust solver among all that I have used.

frederikgeth:
The objective values are definitely not the same between SOC relaxation of SDP in NLP/Ipopt vs SOC/Mosek. The differences should come down to which psd_to_soc function is called [here](https://github.com/frederikgeth/ThreePhasePowerModels.jl/blob/master/src/core/relaxation_scheme.jl).

frederikgeth:
With any other basemva than 1MVA, the results don't match OpenDSS. This isn't expected behaviour, right?

frederikgeth:
That'll be it for this week, and next week I've got a lot of other things to do so progress will be slow. 

Some notes:
- As the new formulations are not defined in PowerModels proper, and as I saw no way to redefine `AbstractWForms`, I had to duplicate some methods. Not sure if this is the clean way to do this
- The reference bus definition is: angles `360/n_conductors` degrees apart, equal magnitude in all phases (but magnitude itself can be optimised)
- There is a slight difference between `SOCBFPowerModel` and `LPdiagUBFPowerModel`/`LPUBFPowerModel` which has something to do with the line shunts. I believe the discrepancy is due to a bug is in `SOCBFPowerModel`. 
- Scaling is to be explored further to understand why NLP-CQCP/Ipopt is different from SOC/Mosek, but I got stuck there as any baseMVA different from 1 (in the OpenDSS file) leads to wrong results.
- The unit tests for the SDP show optimality close to that of the NLP (absolute difference of less than 1 on +/-55 000) for the radial cases. There is a significant cap for the meshed case, but this is expected.
- The tp_pf_bf problem definition is unfinished

Comments welcome.

ccoffrin:
Closes #58 

Progress on #61, OTS still needs to be updated.

PR #85, opened 2018-08-14
Further updating of documentation

frederikgeth:
Main change: clarifying link between mathematical model and function names for BIM and BFM, mirroring https://github.com/lanl-ansi/PowerModels.jl/pull/352

Changelog is updated.


docs/src/formulation-details.md
```diff
@@ -15,12 +15,12 @@ Applying the standard BIM voltage cross-product (sine and cosine) substitution t
 
 ## `SDPUBFForm`
 The BFM SDP relaxation as described in:
-- Gan, L., & Low, S. H. (2014). Convex relaxations and linear approximation for optimal power flow in multiphase radial networks. In PSSC (pp. 1–9). Wroclaw, Poland. https://doi.org/10.1109/PSCC.2014.7038399
+- Gan, L., & Low, S. H. (2014). Convex relaxations and linear approximation for optimal power flow in multiphase radial networks. In PSSC (pp. 1–9). Wroclaw, Poland. <https://doi.org/10.1109/PSCC.2014.7038399>
 Note that this formulation is complex-valued and additional steps are needed to implement this in JuMP.
```
ccoffrin: I think we can get a more maintainable solution to this by putting these links in the docstring when the type is defined, then autodoc will pick them up automatically.  If you want to call out specific doc strings items at this location, there is a way to do that as well.

frederikgeth:
I tried to mirror the approach in https://github.com/lanl-ansi/PowerModels.jl/pull/356, but documenter complains and doesn't render the docstrings correctly:
```
 !! Duplicate docs found for 'SDPUBFPowerModel'. [src/formulation-details.md]
 !! Duplicate docs found for 'SOCNLPUBFPowerModel'. [src/formulation-details.md]
 !! Duplicate docs found for 'SOCConicUBFPowerModel'. [src/formulation-details.md]
 !! Duplicate docs found for 'LPfullUBFPowerModel'. [src/formulation-details.md]
 !! Duplicate docs found for 'LPdiagUBFPowerModel'. [src/formulation-details.md]
 !! Duplicate docs found for 'LPLinUBFPowerModel'. [src/formulation-details.md]
```
I don't think there are duplicates though. I will pick this up again tomorrow or next week.

docs/src/formulation-details.md
```diff
@@ -1,46 +1,64 @@
...
+```
 
-## `StandardACPForm`
-Real-valued formulation from:
-- Formulation without shunts: Mahdad, B., Bouktir, T., & Srairi, K. (2006). A three-phase power flow modelization: a tool for optimal location and control of FACTS devices in unbalanced power systems. In IEEE Industrial Electronics IECON (pp. 2238–2243).
+## `ACPPowerModel`
+Real-valued formulation from (extended with shunts):
+```
+@INPROCEEDINGS{4153495,
+author={B. Mahdad and T. Bouktir and K. Srairi},
```
frederikgeth: @ccoffrin, do you have a better / older reference for this? Preferably one with the shunts already included?

Furthermore, we can't really use a call to 
```
@docs
ACPPowerModel
```
here, as this would collect the reference for the balanced formulation.

docs/src/formulation-details.md
```diff
@@ -1,46 +1,64 @@
...
-## `SOCWRForm`
-Applying the standard BIM voltage cross-product (sine and cosine) substitution tricks to `StandardACPForm` results immediately in a SOC formulation.
 
-## `SDPUBFForm`
-The BFM SDP relaxation as described in:
-- Gan, L., & Low, S. H. (2014). Convex relaxations and linear approximation for optimal power flow in multiphase radial networks. In PSSC (pp. 1–9). Wroclaw, Poland. https://doi.org/10.1109/PSCC.2014.7038399
-Note that this formulation is complex-valued and additional steps are needed to implement this in JuMP.
+## `SDPUBFPowerModel`
+```@docs
+SDPUBFPowerModel
```
frederikgeth: @ccoffrin I don't really understand why this happens:
```
 !! Duplicate docs found for 'SDPUBFPowerModel'. [src/formulation-details.md]
```
It may have something to do with a const and the constructor having the same function name, but I don't understand why it causes problems here but not in PowerModels proper.

frederikgeth: Just tried these two qualified variants:
```
ThreePhasePowerModels.SDPUBFPowerModel
ThreePhasePowerModels.SDPUBFPowerModel()
``` 
Neither solves the issue. 😕

ccoffrin: At this point I think we need to post an issue on Documentor.jl or a question on discourse.

Issue #86, opened 2018-08-17
parse_opendss error when specifying isc1 and isc3


sanderclaeys:
When specifying the isc3 and isc1 property of the circuit in a dss file, the parser fails at line 626 of dss_structs.jl. This error seems to be caused by a typo:

if (haskey(kwargs, :mvasc3) && haskey(kwargs, :mvasc1)) || (haskey(kwargs, :isc3) && haskey(**isc1**, :isc1))

The same typo occurs on line 633.

ccoffrin:
A PR that fixes this issue would be welcome.  In addition to the fix, we will also need a minimal data file that exhibits the issue, so it does not resurface later on.

PR #87, opened 2018-08-17
typo fix in dss_structs.jl

sanderclaeys:
solving issue #86 

corrected two typos (isc1=>kwargs) and added a version of case3_balanced.dss that specifies ISC1 and ISC3 instead of MVAsc1 and MVAsc3, which revealed this bug in the first place

### DIFF
--- a/src/io/dss_structs.jl
+++ b/src/io/dss_structs.jl
@@ -623,14 +623,14 @@ function createVSource(bus1, name::AbstractString, bus2=0; kwargs...)
 
     Zbase = basekv^2 / basemva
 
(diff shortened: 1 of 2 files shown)


ccoffrin:
This looks great.  Thanks!  Will merge it in later today.

ccoffrin:
Note that in adding a test of this network, I had to change the parameter values so that the effect was noticeable in the opf solution, https://github.com/lanl-ansi/ThreePhasePowerModels.jl/commit/1eae4d6f4a5439e061a6fd89e541d1dcc52197b8.

PR #91, opened 2018-10-01
Fix SDP Tests

ccoffrin:
Fixes based on latest version of SCS.

@frederikgeth, the new solver version seems to be much improved over the last one, but some solutions did have notable changes.

Will keep this open for 24h for discussion.

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,7 +1,7 @@
 julia 0.6
 
 Cbc
-SCS
+SCS 0.4
 Ipopt
-Juniper 0.2.0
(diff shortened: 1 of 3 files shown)


frederikgeth:
No objections, this seems like an improvement :)

Sidenote: for the SDP relaxation and the NLP result to be fully consistent in terms of optimality, https://github.com/lanl-ansi/ThreePhasePowerModels.jl/issues/67 needs to be addressed. (the AC form has the freedom to choose the magnitude in each phase of the slackbus independently, the convex formulations don't at the moment).

PR #102, opened 2018-11-13
Add Storage Component Modeling

ccoffrin:
CC @frederikgeth

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@ ThreePhasePowerModels.jl Change Log
 ===================================
 
 ### Staged
+- nothing
+
+### v0.1.3
+- Add opf with multi-phase storage model
 - Add support for parsing OpenDSS Storage objects
 - Minor fix to branch parsing in matlab format
 - Minor fix to OpenDSS parser (parsing ~ lines with preceeding whitespace)


src/prob/tp_test.jl
```diff
@@ -0,0 +1,130 @@
...
+        PMs.constraint_dcline(pm, i, cnd=c)
+    end
+
+    PMs.objective_min_fuel_cost(pm)
+end
+
+
+
+
+"multi-network opf with storage"
```
frederikgeth: "multi-period opf with storage" seems like a more obvious description of the problem? (it uses the multi-network feature)

src/prob/tp_test.jl
```diff
@@ -0,0 +1,130 @@
...
+function run_mn_tp_strg_opf(data::Dict{String,Any}, model_constructor, solver; kwargs...)
+    return PMs.run_generic_model(data, model_constructor, solver, post_mn_tp_strg_opf; multiconductor=true, multinetwork=true, kwargs...)
+end
+
+""
+function post_mn_tp_strg_opf(pm::GenericPowerModel)
+    for (n, network) in nws(pm)
```
frederikgeth: just a suggestion, eventually we could consider the multi-period semantics here as well, e.g.
```
timesteps = nws(pm) #we use the network index to represent time steps in multi-period problems
for (t, timestep) in timesteps
...
```

ccoffrin:
@frederikgeth, you make a good point that also crossed my mind when using multi-networks to implement multi-period OPF problems.  I think it would be good if PowerModels included some helper functions specifically for multi-period models.  At the moment, I don't have any specific ideas for how to model this.  Suggestions, feature requests, and PRs welcome on this topic.

PR #103, opened 2018-11-13
FIX: Case sensitive filesystems for OpenDSS parse

### LINKED ISSUE #101: Case Sensitivity in OpenDSS redirect
The parsing of the redirect command seems to have an error on operating systems where file names are case sensitive.

pseudocubic:
In the case of case-sensitive filesystems, an OpenDSS 'redirect',
'compile', or 'buscoords' command will fail because it will not be able
to find the file. To solve this, `lowercase()` commands were moved to
later in the parse_file function's primary loop.

Closes #101


ccoffrin:
LGTM, merge once CI passes and add a note to the changelog.

PR #104, opened 2018-11-13
ADD: 'source_id' to identify source of PM components from OpenDSS

### LINKED ISSUE #100: Source Names
Lets add a feature to the OpenDSS parser that adds a `source_id` field to each component with the raw OpenDSS name including the component prefix like, `load.abc.1.2`.  This will help users track how the components in OpenDSS translate into the PowerModels data model.

pseudocubic:
Adds three source id related fields to supported PowerModels components,
including "load", "shunt", "branch" (incl. transformers and starbuses), "gen",
"storage", and "pysystem". Buses do not have such fields because of
their implicit declarations in OpenDSS, with the exception of starbuses,
who originate from 3-winding transformers. 

Each supported component will contain three new source-related fields:
"source_id" and "active_phases". "source_id" has the format
"source_type.source_name". "active_phases" is represented as a sparse
array, where each active phase number is included, e.g. `[2, 3]` if phases
2 and 3 are active.

The function `parse_dss_source_id(component)` will return a custom struct
with members `dss_type`, `dss_name` and `active_phases` for easy
aggregation of this information.

Unit tests added for existing models.

Closes #100

### DIFF
--- a/src/io/opendss.jl
+++ b/src/io/opendss.jl
@@ -6,12 +6,16 @@ struct dss_source_id
     dss_type::String
     dss_name::String
     active_phases::Array{Int}
+    id_string::String
 end
 
 
 "Parses a component's OpenDSS source information into the `dss_source_id` struct"
 function parse_dss_source_id(component::Dict)::dss_source_id
-    return dss_source_id(component["source_type"], component["name"], component["active_phases"])
+    dss_type, dss_name = split(component["source_id"], '.')
+    active_phases = component["active_phases"]
+    source_id_str = "$dss_type.$dss_name.$(join(["$n" for n in 1:length(active_phases) if active_phases[n] == 1], '.'))"
+    return dss_source_id(dss_type, dss_name, active_phases, source_id_str)
 end
 
 
@@ -49,8 +53,7 @@ function create_starbus(tppm_data::Dict, transformer::Dict)::Dict
 
     nodes = .+([parse_busname(transformer["buses"][n])[2] for n in length(transformer["buses"])]...)
     starbus["active_phases"] = [nodes[n] > 0 ? 1 : 0 for n in 1:nconductors]
-    starbus["source_name"] = transformer["name"]
-    starbus["source_type"] = "transformer"
+    starbus["source_id"] = "transformer.$(transformer["name"])"
 
     return starbus
 end
@@ -206,8 +209,7 @@ function dss2tppm_load!(tppm_data::Dict, dss_data::Dict, import_all::Bool)
             loadDict["status"] = convert(Int, defaults["enabled"])
 
             loadDict["active_phases"] = [nodes[n] > 0 ? 1 : 0 for n in 1:nconductors]
-            loadDict["source_type"] = "load"
-            loadDict["source_name"] = defaults["name"]
+            loadDict["source_id"] = "load.$(defaults["name"])"
 
             loadDict["index"] = length(tppm_data["load"]) + 1
 
@@ -254,8 +256,7 @@ function dss2tppm_shunt!(tppm_data::Dict, dss_data::Dict, import_all::Bool)
             shuntDict["index"] = length(tppm_data["shunt"]) + 1
 
             shuntDict["active_phases"] = [nodes[n] > 0 ? 1 : 0 for n in 1:nconductors]
-            shuntDict["source_type"] = "capacitor"
-            shuntDict["source_name"] = defaults["name"]
+            shuntDict["source_id"] = "capacitor.$(defaults["name"])"
 
             used = ["bus1", "phases", "name"]
             PMs.import_remaining!(shuntDict, defaults, import_all; exclude=used)
@@ -290,8 +291,7 @@ function dss2tppm_shunt!(tppm_data::Dict, dss_data::Dict, import_all::Bool)
                 shuntDict["index"] = length(tppm_data["shunt"]) + 1
 
                 shuntDict["active_phases"] = [nodes[n] > 0 ? 1 : 0 for n in 1:nconductors]
-                shuntDict["source_type"] = "reactor"
-                shuntDict["source_name"] = defaults["name"]
+                shuntDict["source_id"] = "reactor.$(defaults["name"])"
 
                 used = ["bus1", "phases", "name"]
                 PMs.import_remaining!(shuntDict, defaults, import_all; exclude=used)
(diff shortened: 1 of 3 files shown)


ccoffrin:
Is this feature designed to include the component type and active phases or will those come later?

pseudocubic:
yes, this includes both of those things, e.g. for some load given in opendss by
```
New Load.L1 phases=2 bus1=some_bus.1.2 kw=10 kvar=5
```
`source_id` would be `"load.l1.1.2"`.

src/io/opendss.jl
```diff
@@ -1,6 +1,14 @@
 # OpenDSS parser
 
 
+"parses 'source_id' into Dict containing the 'type', 'name' and active 'phases' of the OpenDSS component"
+function parse_source_id(source_id::String)
```
ccoffrin: probably should be called something like `parse_opendss_source_id`, for cases when `source_id` comes from a non opendss file.

src/io/opendss.jl
```diff
@@ -1,6 +1,14 @@
...
+    component_type, component_name, active_phases = split(source_id, '.', limit=3)
+    nodes = [contains(active_phases, "$n") for n in 1:3]
```
ccoffrin: Instead of a list of booleans it may be easier to have a set of ints.  This will be easier to loop over and do `contains` checks.

src/io/opendss.jl
```diff
@@ -1,6 +1,14 @@
...
+    return Dict("type"=>component_type, "name"=>component_name, "phases"=>nodes)
```
ccoffrin: This may be the exception to our Dict standard, a tuple or struct may be more appropriate.  Can you think of any case where one would want to extend the the fields these three fields?  If not, probably a struct is a better fit.

src/io/opendss.jl
ccoffrin: Also, if you choose to break up the component name and phase info into two separate fields then this function could take as input a component Dict rather than a string.

In that case, probably the name should also be revised.

pseudocubic:
@ccoffrin I have updated my proposed design of this feature, see the updated first comment.

src/io/opendss.jl
```diff
@@ -1,11 +1,21 @@
...
 
 
-"parses 'source_id' into Dict containing the 'type', 'name' and active 'phases' of the OpenDSS component"
-function parse_source_id(source_id::String)
-    component_type, component_name, active_phases = split(source_id, '.', limit=3)
-    nodes = [contains(active_phases, "$n") for n in 1:3]
-    return Dict("type"=>component_type, "name"=>component_name, "phases"=>nodes)
+"Structure representing OpenDSS `dss_source_id` giving the type of the component `dss_type`, its name `dss_name`, and the active phases `active_phases`"
+struct dss_source_id
+    dss_type::String
```
ccoffrin: `AbstractString` may be more flexible.  I recently ran into some type checking bugs with`String` vs `SubString{String}` in Julia v1.0

src/io/opendss.jl
```diff
@@ -1,11 +1,21 @@
...
+    dss_name::String
+    active_phases::Array{Int}
```
ccoffrin: Any reason to prefer an Array{Int} here over a Set{Int}?  Semantically, I think Set is more accurate.

src/io/opendss.jl
```diff
@@ -1,11 +1,21 @@
 # OpenDSS parser
 
 
-"parses 'source_id' into Dict containing the 'type', 'name' and active 'phases' of the OpenDSS component"
-function parse_source_id(source_id::String)
-    component_type, component_name, active_phases = split(source_id, '.', limit=3)
-    nodes = [contains(active_phases, "$n") for n in 1:3]
-    return Dict("type"=>component_type, "name"=>component_name, "phases"=>nodes)
+"Structure representing OpenDSS `dss_source_id` giving the type of the component `dss_type`, its name `dss_name`, and the active phases `active_phases`"
+struct dss_source_id
```
ccoffrin: In Julia camel case is the convention for type names.  So maybe `DSSSourceID`?

src/io/opendss.jl
```diff
@@ -1,11 +1,21 @@
...
+    dss_type::String
+    dss_name::String
+    active_phases::Array{Int}
+    id_string::String
```
ccoffrin: I am not sure what this would be used for.  Isn't it possible to easily reconstruct this field with a function that takes in a `DSSSourceID` and spits out a string?

src/io/opendss.jl
```diff
@@ -41,9 +51,9 @@ function create_starbus(tppm_data::Dict, transformer::Dict)::Dict
     starbus["bus_type"] = 1
     starbus["index"] = starbus_id
 
-    nodes_all = .+([parse_busname(transformer["buses"][n])[2] for n in length(transformer["buses"])]...)
-    active_phases = join(["$n" for n in 1:nconductors if nodes_all[n] > 0], ".")
-    starbus["source_id"] = "transformer.$(transformer["name"]).$(active_phases)"
+    nodes = .+([parse_busname(transformer["buses"][n])[2] for n in length(transformer["buses"])]...)
+    starbus["active_phases"] = [nodes[n] > 0 ? 1 : 0 for n in 1:nconductors]
```
ccoffrin: Instead of a vector of booleans, how about a sparse set of ints, like `Set{Int}([2,3])`.

Later on we can then do things like, `for p in comp["active_phases"] println(pg[p]) end`.

src/io/opendss.jl
```diff
@@ -198,8 +208,8 @@ function dss2tppm_load!(tppm_data::Dict, dss_data::Dict, import_all::Bool)
             loadDict["qd"] = PMs.MultiConductorVector(parse_array(defaults["kvar"] / 1e3, nodes, nconductors))
             loadDict["status"] = convert(Int, defaults["enabled"])
 
-            active_phases = join(["$n" for n in 1:nconductors if nodes[n] > 0], ".")
-            loadDict["source_id"] = "load.$(defaults["name"]).$(active_phases)"
+            loadDict["active_phases"] = [nodes[n] > 0 ? 1 : 0 for n in 1:nconductors]
```
ccoffrin: Here I would recommend a sparse vector of phase ints rather than booleans, i.e. `[2,3]` instead of `[0,1,1]`.

Then we can use it interchangeably with the sets semantics proposed above.

src/io/opendss.jl
pseudocubic: my only concern with using Sets is that they seem to json serialize into lists. If you don't think that would be a problem, then it seems fine with me.

ccoffrin: In the PowerModels data dict we can use a sparse array (i.e. `[2,3]`) for JSON compatibility.  In the struct we can use a Set because it is already a native Julia type.

PR #105, opened 2018-11-14
ADD: OpenDSS PVSystem to PM Generators

### LINKED ISSUE #11: PV Generators
Specification TBD.

pseudocubic:
Adds a section to the `dss2tppm_gen!` function to parse PVSystem OpenDSS objects into simple generators as a first approximation. Uses exclusively the `kVA` property to determine the generator limits `pmin/pmax` and `qmin/qmax`, i.e. `0/kVA` and `-kVA/kVA`, respectively.

Added unit test assumes more generation than necessary from PV, and therefore power is expected to be injected back into the sourcebus generator (transmission network).

There are several supported OpenDSS properties for PVSystems that might be more appropriate, and I welcome discussion on the best practice to get the closest approximation to what OpenDSS might be doing. 

In particular, `Pmpp` (Power at the max power point, kW @ 1 kW/m2 of irradiation) seems like it might be a better value to determine `pmax`. As a first step I was thinking we could assume default values for `P-Tcurve`, `temperature`, and `irradiation`, (i.e. constant, 25C, and 1, respectively), which are all required values to calculate the output power from the PV array. This seems to make `Pmpp` the hard upper limit on PV generation.

Another comment is that in the current implementation of `dss2tppm_pv!`, the OpenDSS properties of `kW` and `kvar` are utilized, but `kw` is undocumented as an input property (although it is indicated that it is an internal value that can be probed), and `kvar` seems to set the PVSystem into a constant kvar mode. Also, I noted that those values are not divided by `nconductors`, as we have done in the other generator cases. I therefore have some confusion about the current implementation of `pvsystem` parsing. 

CC @frederikgeth @ccoffrin 

Closes #11

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ ThreePhasePowerModels.jl Change Log
 - Minor fix to OpenDSS parser (parsing ~ lines with preceeding whitespace)
 - Fixed parsing OpenDSS files containing redirect/compile/buscoords on case-sensitive filesystems
 - Add 'source_id' field to components parsed from OpenDSS, to help determine origin and active phases
+- Add conversion of OpenDSS PVSystem objects into generators, using KVA for generator limits
 
 ### v0.1.2
 - Add support for network flow approximation formulation, NFAPowerModel
--- a/src/io/opendss.jl
+++ b/src/io/opendss.jl
@@ -416,6 +416,52 @@ function dss2tppm_gen!(tppm_data::Dict, dss_data::Dict, import_all::Bool)
             push!(tppm_data["gen"], genDict)
         end
     end
+
+    if haskey(dss_data, "pvsystem")
+        for pv in dss_data["pvsystem"]
+            if haskey(pv, "like")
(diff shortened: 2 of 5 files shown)


frederikgeth:
Sorry for the delay. This is an awesome addition! 
I just ran one of our internal OpenDSS test cases with numerous single-phase PVSystem objects (e.g. `New PVSystem.SOLAR65 Phases=1 Bus1=192.2 model=1 kVA=1 kV=0.24 irradiance=1 pmpp=1 kvar=1 %cutin=0 %cutout=0 temperature=25
`) and it returned reasonable results.

The logging messages of the parser could be updated that some value translation from PVSystem objects to generators actually occurs. A test case with a single-phase PV system may be useful as well.

ccoffrin:
@pseudocubic, you've tagged this for Release v0.2.0, are there any breaking changes in here?  If not we can put it in v0.1.3.

pseudocubic:
@ccoffrin no breaking changes, i only put it in v0.2.0 because we had discussed previously that it was not blocking for tagging v0.1.3

ccoffrin:
Closes #11

PR #110, opened 2018-12-15
Updates for PowerModels v0.9

ccoffrin:
The only thing of particular note is that I fixed a bug in the Matlab and OpenDSS parsers where some empty arrays of optional components were note getting turned into empty dicts.


ccoffrin:
Most likely the new support for linear objective functions was the source of this difference.  I did not look into it in much detail, because the solution to reactive power is likely not unique in that model.  Small perturbations in model can result in different solutions.

PR #115, opened 2019-03-12
zero series impedance branches

sanderclaeys:
The transformer loss model is easier to implement if pi-sections with zero series impedance are supported (which are still allowed to contain shunts, and are therefore not lossless). 

In the constraint template of constraint_ohms_tp_yt_from (and to), I check whether all elements of the series impedance are below a threshold (default 1E-13). This threshold is needed when the impedance matrix is calculated from symmetrical components. This transformation introduces some numerical errors, causing a branch with exactly zero impedance in series components, to have small impedance values in the abc frame (in the order of 1E-14).

If all impedance values are below this threshold, a new constraint is applied, constraint_ohms_tp_yt_from_impzero.

I also added two unit tests for different formulations. They check whether the objective value remains the same after adding a zero impedance branch to the network.

### DIFF
--- a/src/core/constraint_template.jl
+++ b/src/core/constraint_template.jl
@@ -35,25 +35,31 @@ end
 
 
 ""
-function constraint_ohms_tp_yt_from(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw, cnd::Int=pm.ccnd)
+function constraint_ohms_tp_yt_from(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw, cnd::Int=pm.ccnd, atol_impzero=1E-13)
     branch = ref(pm, nw, :branch, i)
     f_bus = branch["f_bus"]
     t_bus = branch["t_bus"]
     f_idx = (i, f_bus, t_bus)
     t_idx = (i, t_bus, f_bus)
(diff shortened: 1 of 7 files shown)


sanderclaeys:
On my laptop, only two tests related to 'test dc polar opf', failed, both on my branch and on master.

sanderclaeys:
I discovered an issue which got past the unit tests; the voltages are not linked in the acp and wr formulation. I will commit a fix tomorrow.

ccoffrin:
Thanks for the PR!  With out knowing the context or motivations for this feature, I will remark that my past experience has been that zero impedance branches lead to more modeling challenges than solutions.  So before merging I would like to discuss if other options are possible options and better understand why this would be the best one, but please feel free to keep this PR open for code comparison and discussions.

sanderclaeys:
Thanks for the feedback. I thought it might be easier afterwards to have this feature as a separate PR rather than mixed in with the transformer models. The first transformer PR should be ready within 1-2 weeks, I think it will be easier to discuss the need for this one after that. I will keep this PR open until then.

ccoffrin:
Sounds good to me.  Thanks!

PR #117, opened 2019-03-27
transformer extension

### LINKED ISSUE #7: Simple Tranformers
Specification TBD.

### LINKED ISSUE #114: Issue with Zb calculation
cc @risjain

If `baseMVA` is three phase, then it should be divided by three. I believe this line should be modified.

https://github.com/lanl-ansi/ThreePhasePowerModels.jl/blob/848b6e9c56011ab0e22f4229a50cda008a8e568a/src/io/opendss.jl#L527

This is the equation to calculate Zbase:

```
Zbase = (basekv / sqrt(3))^2 / baseMVA_perphase # OR

Zbase = (basekv / sqrt(3))^2 / ( baseMVA_nphase / nconductors ) # i.e.

Zbase = (basekv / sqrt(3))^2  * nconductors / ( baseMVA_nphase )
```

If you remove the ` * nconductors` subexpression, you'll get the right values for rmatrix, xmatrix in per unit without multiplying by 3 in the following lines:

https://github.com/lanl-ansi/ThreePhasePowerModels.jl/blob/848b6e9c56011ab0e22f4229a50cda008a8e568a/src/io/opendss.jl#L524-L525

### LINKED ISSUE #116: pi-model shunt overflow and factor 3 missing
There is a factor (1/3) missing in the opendss parser for the pi-model shunts. The rmatrix and xmatrix variables are multiplied by 3 to compensate for the voltage base (as discussed in #114). However, in a similar fashion cmatrix should be divided by 3. This probably gets past the unit tests, because in many lines the effect of the shunt capacitance is quite small. In my code, setting the shunt to zero only affects the voltage magnitude by 1E-5 in p.u.

Whilst validating this, I ran into another issue. A large shunt capacitance is needed to see the effect discussed above. Whilst increasing the value of C1 and C0, I ran into overflow issues. The implementation of createLine in dss_structs.jl uses arrays of Float16 precision in many places. Changing the precision to Float64 solves this issue.

sanderclaeys:
This PR adds support for parsing OpenDSS transformers into TPPM. They are represented as compositions of ideal two-winding transformers, and the loss model is mapped to a network of branches. The loss model can potentially be initialized with branches which have zero series impedance, but an additional algorithm simplifies it, removing these branches.

Furthermore, an algorithm is added which sets a voltage base for each bus in the network, and converts branches and shunts to be consistent with this new base. The voltage base changes across transformers as the ratio of the nominal voltage of the windings. Thanks to this, 1 p.u. is still a good guess for the initial point (if the transformer ratings are consistent with the voltage of the slack bus).

An ideal two-winding transformer is a combination of a (possibly variable) tap-changer and a winding connection configuration (delta, wye and all vector groups). The pf problem formulation always considers the taps to be fixed, and represents them as parameters. The opf adds the taps as variables, as long as not all of them are fixed (per transformer). Should the opf with variable tap be moved to a new problem formulation, whilst fixing the taps in the current opf problem formulation?

Several unit tests were added, and the opendss parser unit test was updated to reflect the changes.

I also attached a pdf with more explanation on why each file was edited. I hope this will make it easier to review this code.
[summary_changes.pdf](https://github.com/lanl-ansi/ThreePhasePowerModels.jl/files/3011184/changes.pdf)

Closes #116 
Closes #114 
Closes #7


pseudocubic:
@sanderclaeys thanks very much for the PR, because this feature changes so much it might take a little longer for us to act on it, but I wanted you to know I've seen it and will be actively looking at it. 

As for the failing tests, these I believe can be fixed by changing the solver to ipopt_solver, instead of using IpoptSolver() with default options. There is a fix for this in our julia-v1 branch, but since we haven't pulled the trigger on dropping support for julia < 1.0, it hasn't made its way to master yet.

sanderclaeys:
@pseudocubic don't hesistate to contact me whilst reviewing it; I will regularly check this thread. I am currently in Australia, so it might take me a while to get back to you because of the time difference.

frederikgeth:
Thanks, @sanderclaeys, great work.

Linking in this issue as well: https://github.com/lanl-ansi/ThreePhasePowerModels.jl/issues/7

src/io/common.jl
```diff
@@ -45,10 +45,18 @@ function check_network_data(data::Dict{String,Any})
...
     return degrees - 360*floor.((degrees .+ 180)/360)
 end
 
+function wraptopi(degrees::MultiConductorVector)
+    MultiConductorVector(wraptopi(degrees[:]))
+end
+
 function wraptopi(radians)
     return radians - 2*pi*floor.((radians .+ pi)/(2*pi))
 end
```
pseudocubic: If the original `wrapto180` and `wraptopi` aren't returning `MultiConductorVector`s, we should ensure that this is working in PowerModels.jl. This could be a problem somewhere else too.

src/io/opendss.jl
```diff
@@ -523,7 +523,7 @@ function dss2tppm_branch!(tppm_data::Dict, dss_data::Dict, import_all::Bool)
 
             rmatrix = parse_matrix(defaults["rmatrix"], nodes, nconductors) * 3  # why?
             xmatrix = parse_matrix(defaults["xmatrix"], nodes, nconductors) * 3  # why?
-            cmatrix = parse_matrix(defaults["cmatrix"], nodes, nconductors)
+            cmatrix = parse_matrix(defaults["cmatrix"], nodes, nconductors) / 3
```
pseudocubic: Since we are now aware of the origin of this factor, could you update the line comments to reflect this?

src/io/opendss.jl
```diff
@@ -968,22 +1418,36 @@ function parse_opendss(dss_data::Dict; import_all::Bool=false, vmin::Float64=0.9
...
+        if isa(tppm_data[key], Array)
+            tppm_data[key] = Dict{String, Any}()
+        end
+    end
 
     for optional in ["dcline", "load", "shunt", "storage", "pvsystem"]
         if length(tppm_data[optional]) == 0
             tppm_data[optional] = Dict{String,Any}()
         end
     end
```
pseudocubic: This loop to turn empty `branch` arrays into a dict can be combined with the loop on line 1437, just add branch to the `optional` list.

src/io/opendss.jl
```diff
@@ -968,22 +1418,36 @@ function parse_opendss(dss_data::Dict; import_all::Bool=false, vmin::Float64=0.9
...
 
+    if haskey(tppm_data, "trans_comp")
+        # this has to be done before calling adjust_sourcegen_bounds!
+        decompose_transformers!(tppm_data)
+        adjust_base!(tppm_data)
+    end
```
pseudocubic: It looks like `trans_comp` is left in place, but is it used later or only during this parsing process? If it is not used I think after the parsing it should be deleted.

src/io/opendss.jl
```diff
@@ -590,108 +589,174 @@ function dss2tppm_transformer!(tppm_data::Dict, dss_data::Dict, import_all::Bool
...
+                # and %rs map to values for the winding series resistance.
+                # Both are set to a (possibly default) value, so it is
+                # impossible to know at this point which was actually specified
+                # by the user. %rs is more general, so %loadloss is not
+                # supported for now.
+                warn(LOGGER, "The %loadloss property is ignored for now.")
+                #TODO handle neutral impedance
+                # neutral impedance is ignored for now; all transformers are
+                # grounded (that is, those with a wye and zig-zag winding).
+                warn(LOGGER, "The neutral impedance, (rg and xg properties), is ignored; the neutral (for wye and zig-zag windings) is connected directly to the ground.")
```
pseudocubic: Tests should be added for these warnings

test/opendss.jl
```diff
@@ -95,10 +92,17 @@ TESTLOG = getlogger(PowerModels)
 
         @test tppm["name"] == "test2"
 
-        @test length(tppm) == 18
+        @test length(tppm) == 20 # 2 more entries for transformer dicts
```
pseudocubic: Might need to be updated if we drop `trans_comp`?

src/io/opendss.jl
```diff
@@ -590,108 +589,174 @@ function dss2tppm_transformer!(tppm_data::Dict, dss_data::Dict, import_all::Bool
...
+                # All of the code is compatible with any number of windings,
+                # except for the parsing of the loss model (the pair-wise reactance)
+                error(LOGGER, "For now parsing of xscarray is not supported. At most 3 windings are allowed, not $nrw.")
+            end
 
-                    transDict["transformer"] = true
-                    transDict["switch"] = false
+            transDict = Dict{String,Any}()
+            transDict["name"] = defaults["name"]
+            transDict["source_id"] = "transformer.$(defaults["name"])"
```
pseudocubic: Missing `active_phases`?

src/core/variable.jl
```diff
@@ -133,3 +133,69 @@ function variable_reactive_bus_power_slack(pm::GenericPowerModel; nw::Int=pm.cnw
         start = PMs.getval(ref(pm, nw, :bus, i), "q_slack_start", cnd)
     )
 end
+
+"generates variables for both `active` and `reactive` power flow at each transformer"
+function variable_tp_trans_flow(pm::GenericPowerModel; kwargs...)
+    variable_tp_trans_active_flow(pm; kwargs...)
+    variable_tp_trans_reactive_flow(pm; kwargs...)
+end
+function variable_tp_trans_active_flow(pm::GenericPowerModel; nw::Int=pm.cnw, bounded=true)
```
pseudocubic: I have been trying to roughly enforce python pep8 conventions in TPPMs especially for function declarations, i.e. two lines between functions, and a docstring for each function. This just helps me with readability and easier debugging later. Docstrings don't all need to be complicated, just enough that I can easily understand the purpose of the function. Please check other functions in this file and others as well (also I noticed this issue in opendss.jl), I didn't mark each one.

src/io/common.jl
pseudocubic: I cannot confirm that this is a problem in PowerModels, these functions both return MultiConductorVectors in their original incarnations. Unless you have an example where it fails I think we can remove the two new function declarations.

src/io/opendss.jl
pseudocubic: After re-reading the notes you sent by email, I think `trans_comp` should be retained only if `import_all` is enabled.

src/io/common.jl
sanderclaeys: This was a quick fix for another issue, that I am only now fully aware of when looking at it again. Take for example the following vector and operations
`a = MultiConductorVector(ones(Float64, 3))`
`floor.(a) .+ a`

This all works as expected. However, the solution dictionary contains vectors of the type
`va = MultiConductorVector(ones(Real, 3))`

Element-wise operations are not defined for the element type Real
`floor.(va) .+ va`

@ccoffrin is this the intended behaviour?

src/io/opendss.jl
sanderclaeys: One potential use for the trans_comp data would be to create a composed transformer in the solution dictionary.  There the solutions could be shown for the entire transformer, and not dispersed over several other dicts. If there is no reference to the compound elements in the data model,  such a transformation would not make any sense. This feature is low priority for us, but we would like to keep the trans_comp dict around for the future.

I implemented it as you proposed; if import_all is not flagged, trans_comp is removed after the call to decompose_transformers.

src/io/opendss.jl
sanderclaeys: Because of the delta-winding, it is not trivial to support 1 and 2 phase transformers. Supporting such transformers will require rethinking the data model.

Instead, if an option other than {busx, busx.1.2.3, busx.0.1.2.3} is specified for a transformer winding, a warning is thrown. The warning states that only three-phase transformers are allowed, and that the bus is considered to be simply busx. A unit test was added to verify that this warning is thrown.

src/core/variable.jl
sanderclaeys: If some of the explanations are still unclear, please let me know.

sanderclaeys:
If you don't mind, I prefer postponing updating the documentation for now. We are planning a few more PRs, and it will be easier to create a coherent explanation once these have all been merged.

src/io/opendss.jl
pseudocubic: Your explanation makes sense, but because `active_phases` is an expected field, like `source_id`, I think you could just populate it with a standard `transDict["active_phases"] = [1, 2, 3]` until 1 or 2 phase transformers are supported and you need to change it.

src/core/variable.jl
pseudocubic: Thanks for that. I did see a couple of functions that you missed: `adjust_base_rec!` and `add_shunt!` in `opendss.jl`

pseudocubic:
Your point about waiting for documentation seems fair given the circumstances, I will create a new Issue as a reminder after this is merged.

src/io/common.jl
pseudocubic: Ok, I managed to track this bug down, it has to do with how broadcasting is implemented in Julia `> v"0.7.0-"`. I've made a PR for this fix in PowerModels, which will also hopefully fix the problem with `norm` in the Julia v0.6 travis tests.

sanderclaeys: Great! The broadcasting magic went beyond my current understanding of Julia. I will remove the wrapper functions in common.jl in the meanwhile.

sanderclaeys: Note that this will break most of the tests in transformer.jl

src/io/opendss.jl
sanderclaeys: I added it to both the transformers in "trans_comp" en  "trans"

src/io/common.jl
pseudocubic: That's okay, once PowerModels gets tagged with these broadcast fixes (@ccoffrin ), I will restart the travis runs, and make sure everything is really working

pseudocubic:
@sanderclaeys Everything is looking good now, once we have a new version of PowerModels with the broadcast fixes and we can ensure the travis tests are passing, I think we'll be ready to merge

ccoffrin:
I would like to take a pass before we merge.  I'll try to get to it tomorrow.  If not, over the weekend.

sanderclaeys:
To be consistent with other components, I removed the check whether "trans" is defined and made sure it is always present.

The transformer version of KCL is now always called, so the ACR, DCP and WR version had to be added for the unit tests to keep on working.

I added a workaround for the MultiConductorVector issue in the unit test, that will keep on working with the fix. Because the current KCL version does not yet include storage, all unit tests pass on my laptop in this version.

src/prob/tp_opf.jl
```diff
@@ -29,14 +29,18 @@ function post_tp_opf(pm::GenericPowerModel)
         PMs.variable_dcline_flow(pm, cnd=c)
     end
 
+    add_arcs_trans!(pm)
```
ccoffrin: All `ref` extensions should come at the top of the model, before variables, constraint, or the objective.

src/prob/tp_opf.jl
```diff
@@ -29,14 +29,18 @@ function post_tp_opf(pm::GenericPowerModel)
...
+    variable_tp_trans_flow(pm)
+    variable_tp_trans_tap(pm)
```
ccoffrin: Please add a separate problem formulation for OLTC.

src/prob/tp_opf.jl
```diff
@@ -55,5 +59,18 @@ function post_tp_opf(pm::GenericPowerModel)
         PMs.constraint_dcline(pm, i, cnd=c)
     end
 
+    if haskey(ref(pm), :trans)
```
ccoffrin: Trans `trans` should be required in the data model (even if empty) so that check can be avoided.

src/prob/tp_pf.jl
```diff
@@ -36,7 +36,13 @@ function post_tp_pf(pm::GenericPowerModel)
         PMs.variable_dcline_flow(pm, bounded=false, cnd=c)
     end
 
+    add_arcs_trans!(pm)
```
ccoffrin: same note as in OPF.

src/prob/tp_pf.jl
```diff
@@ -36,7 +36,13 @@ function post_tp_pf(pm::GenericPowerModel)
...
+    variable_tp_trans_flow(pm, bounded=false)
+
     constraint_tp_voltage(pm)
+    # lower bound is often needed to converge
+    # this constraint only sets bound if previously unset
+    constraint_tp_voltage_mag_unbound(pm, vmin=0.3, vmax=Inf)
```
ccoffrin: I suggest doing this as a data pre-processing at the top of this formulation.

src/prob/tp_pf.jl
```diff
@@ -83,4 +89,13 @@ function post_tp_pf(pm::GenericPowerModel)
             PMs.constraint_voltage_magnitude_setpoint(pm, t_bus["index"], cnd=c)
         end
     end
+
+    if haskey(ref(pm), :trans)
```
ccoffrin: same note as OPF.

src/prob/tp_opf.jl
sanderclaeys: I added this check at the end of the OpenDSS parser. I also added this check in add_arcs_trans!, because otherwise data being parsed from the matlab format would not contain the trans key.

src/form/acp.jl
```diff
@@ -133,3 +167,128 @@ function constraint_ohms_tp_yt_to_on_off(pm::GenericPowerModel{T}, n::Int, c::In
...
+                    1/vm_im[c]*(p_to[c]*sin(va_im[c])-q_to[c]*cos(va_im[c])) # i_to_im[c]
+              for c in 1:ncnd)
+            == 0
+        )
+    end
+end
+
+
+"Links the power flowing into both windings of a variable tap transformer."
+function constraint_tp_trans_flow_var(pm::GenericPowerModel, i::Int, f_bus::Int, t_bus::Int, f_idx, t_idx, Ti_fr, Ti_im; nw::Int=pm.cnw)
```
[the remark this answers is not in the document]
sanderclaeys: Gladly. I only did it this way to mimic the branch model, but personally I prefer to lump it into one constraint.

src/form/acp.jl
```diff
@@ -133,3 +167,128 @@ function constraint_ohms_tp_yt_to_on_off(pm::GenericPowerModel{T}, n::Int, c::In
...
+            @constraint(pm.model, var(pm, nw, c, :tap)[i]==tapset[c])
+        end
+    end
+end
+
+
+"Links the power flowing into both windings of a fixed tap transformer."
+function constraint_tp_trans_flow(pm::GenericPowerModel, i::Int, f_bus::Int, t_bus::Int, f_idx, t_idx, Ti_fr, Ti_im; nw::Int=pm.cnw)
+    # the intermediate bus voltage is saved as an expression
+    ncnd  = 3
```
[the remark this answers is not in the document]
sanderclaeys: I will add a check in the new constraint that will lump together all transformer constraints.

src/core/variable.jl
```diff
@@ -126,10 +128,92 @@ function variable_active_bus_power_slack(pm::GenericPowerModel; nw::Int=pm.cnw,
...
+            basename="$(nw)_$(cnd)_p_trans",
+            start=0
+        )
+        if bounded
+            for arc in ref(pm, nw, :arcs_trans)
+                tr_id = arc[1]
+                flow_lb  = -ref(pm, nw, :trans, tr_id, "rate_a")[cnd]
+                flow_ub  =  ref(pm, nw, :trans, tr_id, "rate_a")[cnd]
+                PMs.setlowerbound(var(pm, nw, cnd, :p_trans, arc), flow_lb)
+                PMs.setupperbound(var(pm, nw, cnd, :p_trans, arc), flow_ub)
```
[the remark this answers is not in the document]
sanderclaeys: I converted these to JuMP functions.
@frederikgeth these functions are used in the branch flow variables as well, I followed those as a template for setting the bounds.

src/core/constraint_template.jl
```diff
@@ -201,3 +219,81 @@ function constraint_tp_storage_exchange(pm::GenericPowerModel, i::Int; nw::Int=p
...
+    # fix the taps with a constraint which are not free
+    # at least one tap will be free
+    tapfix = trans["tapfix"]
+    tapset = trans["tapset"]
+    constraint_tp_trans_tap_fix(pm, i, tapfix, tapset)
+end
+
+
+"Links the power flowing into both windings of a fixed tap transformer."
+function constraint_tp_trans_flow(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw)
```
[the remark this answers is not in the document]
sanderclaeys: I totally agree. I only did it this way to mimic the branch model, but personally I prefer to lump it into one constraint.

src/core/constraint_template.jl
```diff
@@ -201,3 +219,81 @@ function constraint_tp_storage_exchange(pm::GenericPowerModel, i::Int; nw::Int=p
...
+    # TODO add checks to simplify
+    trans = ref(pm, :trans, i)
+    f_bus = trans["f_bus"]
+    t_bus = trans["t_bus"]
+    constraint_tp_trans_voltage_var(pm, i, f_bus, t_bus, Tv_fr, Tv_im, Cv_to)
+    # fix the taps with a constraint which are not free
+    # at least one tap will be free
+    tapfix = trans["tapfix"]
+    tapset = trans["tapset"]
+    constraint_tp_trans_tap_fix(pm, i, tapfix, tapset)
```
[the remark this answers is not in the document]
sanderclaeys: Agreed. Proposed changes:
- In `tp_pf` and `tp_opf`, all transformers are treated as non-dispatchable components (labelled as trans).
- In `tp_opf_oltc`, all transformers are treated as dispatchable components (labelled as oltc).

So if I understand correctly the behaviour of the PR in PowerModels, a transformer in the data model would be split over two components in the reference dict, oltc and oltc_const/trans? Given the above changes, this would be easy to integrate later on.

src/core/variable.jl
```diff
@@ -126,10 +128,92 @@ function variable_active_bus_power_slack(pm::GenericPowerModel; nw::Int=pm.cnw,
...
+        end
+    end
+end
+
+
+"""
+Create tap variables;
+only do this for OLTCs which have at least one non-fixed tap.
+"""
+function variable_tp_trans_tap(pm::GenericPowerModel; nw=pm.cnw, kwargs...)
```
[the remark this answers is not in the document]
sanderclaeys: Yes, it is. See discussion on 'dispatchable' functionality

src/form/acp.jl
```diff
@@ -133,3 +167,128 @@ function constraint_ohms_tp_yt_to_on_off(pm::GenericPowerModel{T}, n::Int, c::In
                                         sum(-b[c,d]*vm_to[c]*vm_fr[d]*cos(va_to[c]-va_fr[d]) +
                                              g[c,d]*vm_to[c]*vm_fr[d]*sin(va_to[c]-va_fr[d]) for d in PMs.conductor_ids(pm))) )
 end
+
+
+"Links the voltage at both windings of a fixed tap transformer."
+function constraint_tp_trans_voltage(pm::GenericPowerModel, i::Int, f_bus::Int, t_bus::Int, tapset::MultiConductorVector, Tv_fr, Tv_im, Cv_to; nw::Int=pm.cnw)
+    ncnd  = 3
```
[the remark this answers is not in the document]
sanderclaeys: Check added at constraint_template level, error thrown if not true.

src/core/constraint_template.jl
```diff
@@ -201,3 +219,81 @@ function constraint_tp_storage_exchange(pm::GenericPowerModel, i::Int; nw::Int=p
     PMs.constraint_storage_complementarity(pm, nw, i)
     constraint_tp_storage_loss(pm, nw, i, storage["storage_bus"], storage["r"], storage["x"], storage["standby_loss"])
 end
+
+
+"Links the voltage at both windings of a fixed tap transformer."
+function constraint_tp_trans_voltage(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw)
```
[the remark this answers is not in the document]
sanderclaeys: The constraints are split in two parts: constraints on the voltage, and constraints on the flow (on both power and current depending on the formulation). These could in principle all be merged in one constraint. But since the voltage constraints can be imposed without reference to the other variables, we prefer this separation.

src/core/constraint_template.jl
```diff
@@ -201,3 +219,81 @@ function constraint_tp_storage_exchange(pm::GenericPowerModel, i::Int; nw::Int=p
...
+    # TODO add checks to simplify
+    f_bus = ref(pm, :trans, i)["f_bus"]
+    t_bus = ref(pm, :trans, i)["t_bus"]
+    tapset = ref(pm, :trans, i)["tapset"]
+    constraint_tp_trans_voltage(pm, i, f_bus, t_bus, tapset, Tv_fr, Tv_im, Cv_to)
+end
+
+
+"Links the voltage at both windings of a variable tap transformer."
+function constraint_tp_trans_voltage_var(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw)
```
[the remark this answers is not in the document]
sanderclaeys: See comment above. I will add an explanation below of the new structure.

src/core/constraint_template.jl
```diff
@@ -34,6 +34,23 @@ function constraint_tp_voltage(pm::GenericPowerModel; nw::Int=pm.cnw)
...
+        constraint_tp_voltage_mag_unbound(pm, bus_id, vmin, vmax, nw=nw)
+    end
+end
+
+
+"""
+In the general case, do not add the bounds;
+expand this on a per formulation case as deemed necessary.
+"""
+function constraint_tp_voltage_mag_unbound(pm::GenericPowerModel{T}, i::Int, vmin::Float64, vmax::Float64; nw::Int=pm.cnw) where T
```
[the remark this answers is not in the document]
sanderclaeys: In the past it was needed sometimes to obtain convergence. The unit tests seem to converge without it, so I simply removed it from the PR without breaking anything.

src/core/constraint_template.jl
```diff
@@ -34,6 +34,23 @@ function constraint_tp_voltage(pm::GenericPowerModel; nw::Int=pm.cnw)
 end
 
 
+"Set loose voltage bounds; they will only be set if the voltage magnitude was unconstrained before."
+function constraint_tp_voltage_mag_unbound(pm::GenericPowerModel; nw::Int=pm.cnw, vmin=0.5, vmax=Inf)
+    for bus_id in ids(pm, pm.cnw, :bus)
```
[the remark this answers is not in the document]
sanderclaeys: Agreed. It would be useful to have some utility functions, where the user can transform a data model by setting standard voltage bounds and so on. Like that, it is also made explicit to the user and should improve awareness of what is going on.

src/prob/tp_pf.jl
sanderclaeys: Removed from the PR; unit tests converge without.

src/form/acp.jl
```diff
@@ -133,3 +167,128 @@ function constraint_ohms_tp_yt_to_on_off(pm::GenericPowerModel{T}, n::Int, c::In
...
+    # intermediate bus voltage, for now ignore tap changer
+    vm_to = [var(pm, nw, c, :vm, t_bus) for c in 1:ncnd]
+    vm_im = @NLexpression(pm.model, [c in 1:ncnd], vm_to[c]*tapset[c]*Cv_to)
```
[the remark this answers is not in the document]
sanderclaeys: I pushed it down into the NLconstraint, both for trans and oltc.

This was a side-effect of trying to reuse code that is shared between a fixed transformer and an OLTC. In the new structure I completely separated the OLTC and a standard transformer. This leads to some duplication of code, but will be more flexible and ready for the 'dispactchable' parameter.

src/form/acp.jl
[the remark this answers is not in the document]
sanderclaeys: In the end I separated completely trans and oltc. Especially for this constraint there is a lot of duplication, but I believe it is more readable like this. As a consequence, there is no longer a call to another constraint here.

src/core/ref.jl
```diff
@@ -34,3 +34,75 @@ function find_ref_buses(pm::GenericPowerModel, nw)
...
+    elseif dyz=='z'
+        #TODO zif-zag here
+    else
+        error(LOGGER, "Illegal connection specification $dyz.")
+    end
+    #Tv_fr = diagm(0=>ones(Float64, 3))*Tminn
+    Cv_to = trans["vnom_kv"][1]*vmult/trans["vnom_kv"][2]
+    # make equations dimensionless
+    bkv_fr = ref(pm, nw, :bus, trans["f_bus"], "base_kv")
+    bkv_to = ref(pm, nw, :bus, trans["t_bus"], "base_kv")
```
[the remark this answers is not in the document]
sanderclaeys: **wye transformer**
This ensures that the equations are always correct, even when the voltage base across a transformer does not scale proportionally to nominal tap ratio.
If we assume that the voltage base is scaled as such, this can be left out.

**delta transformer**
There is a tricky subtlety for a delta transformer. 
Say that the nominal voltage at the primary is V1, and at the secondary is V2. The nominal tap ratio then is V1:V2=V1/V2
However, a delta winding boosts up the voltage by a factor of sqrt(3). To obtain the nominal voltage at the secondary winding, the physical tap ratio (which should go into the equations) is a factor sqrt(3) higher, V1.sqrt(3) : V2 
Schematically:
V1---(delta-to-wye)---V1.sqrt(3)---N:1---V2
This is expressed by vmult in the code above.

sanderclaeys:
Thanks for the elaborate feedback! I reorganized the code quite some in addressing these comments.

The mathematical modelling for transformers and OLTCs is now completely separate. They still share the same entries in the data model, but the distinction is made by the problem definition.

Like this, it will be easy in the future to map
- transformers -> non-dispatchable transformers
- oltcs -> dispatchable transformers

The one thing left to do (before further feedback at least) is rewriting `calc_tp_trans_Tvi` to reflect the changes, over dicts and not ref. Also some docstrings have to be added.

I will be traveling the next days with limited connectivity, but will resume work on this next week Tuesday (Australia).

ccoffrin:
Great thanks!  The changes are generally looking good to me.  

Can you point me to a description of the transformer component data model that is used inside of PowerModels?  So I can review its design as well?

sanderclaeys:
The blue properties could default to the values shown in the example. The red properties should always be set explicitly.
Screenshot 2019-04-19 at 09 05 39

ccoffrin:
Ok thanks.  My primary interest is in the data spec of the simplified transformer component inside of power models.

sanderclaeys:
| name | example/default | unit | comment |
|---|---|---|---|
| buses | [1,2] | N/A | 2 values, primary and secondary respectively |
| conn | 123+y | N/A | values: {d,y,z}x{-,+}x{123, 312, 31} |
| vnom_kv | [1.0, 1.0] | kv | 2 values, primary and secondary respectively |
| tapset | [1.0, 1.0, 1.0] | vnom_kv[w] | 3 values, one per phase |
| tapstep | [1/16, 1/16, 1/16] | vnom_kv[w] | discrete step size (not used in math. model) |
| tapmin | [0.9, 0.9, 0.9] | vnom_kv[w] | 3 values, one per phase |
| tapmax | [1.1, 1.1, 1.1] | vnom_kv[w] | 3 values, one per phase |
| tapfixed | [true, true, true] | boolean | 3 values, one per phase |

ccoffrin:
Thanks for the data model doc.  

For context, we will have two types of components, a generic n-winding transformer and a more restricted 2-winding transformer model.   Based on your example, I assume we are talking about the 2-winding case here.

These are my thoughts.

The connection points should be `f_bus` and `t_bus` for consistency with branch naming.  Similarly it should be `vnom_kv_fr` `vnom_kv_to`.   

`vnom_kv` sounds redundant, how about `vm_nom` (units don't need be explicit in the name)?  Might also consider `nom_kv`. 

`tapset`->`tm`, tapmin =>`tm_min`, `tapmax` =>`tm_max`, `tapfixed` => `tm_fixed` to be consistent with other components.

I am thinking `conn` should be called something like `config` or `conn_type`.  I think the data should be an array of three string values.  The order of the strings should be setup so that concatenating them in order results in a larger string that is intuitive to the user.

It seems like this component is implicitly 3-phases.  How does that interact with the multi-phase / multi-conductor generalization?  I am fine to focus only on the 3-phase/3-conductor and 3-phase/4-conductor cases in the near term.  But it is preferable if we know how the data model would support the more general cases in the future.

sanderclaeys:
At the lowest level, there is only one component, the two-winding transformer. A n-winding transformer is always mapped to a set of two-winding transformers. That means that it is only possible right now to specify n-winding transformers in OpenDSS directly. Long-term it would be good to have a higher-level data model in TPPM, with components that are automatically decomposed to lower-level components (in this example, a n-winding transformer to a set of branches and two-winding transformers).

I agree with the name changes, since it is now explicitly a two-terminal component.

To have a more general two-winding transformer, I propose the following data model. This could accommodate n-phase systems, if desired at some point.

For example, take a 5 conductor system. A 4 phase delta wye transformer could be specified as
config_fr: delta
cnd_fr: [1,2,3,4]
config_to: wye
cnd_to: [1,2,3,4,5]

| name | example/default | unit | comment |
|---|---|---|---|
| f_bus | 1 | N/A |  |
| t_bus | 1 | N/A |  |
| config_fr | wye | N/A | wye\|delta |
| config_to | wye | N/A | assumed to be wye for now |
| cnd_fr | [1,2,3] | N/A |  |
| cnd_to | [1,2,3] | N/A | assumed to be [1,2,3] for now |
| vm_nom_fr | 1.0 | kv |  |
| vm_nom_to | 1.0 | kv |  |
| tm | [1.0, 1.0, 1.0] | vnom_kv[1]/vnom_kv[2] | 3 values, one per phase |
| tm_step | [1/16, 1/16, 1/16] | vm_nom_fr/vm_nom_to | discrete step size (not used in math. model) |
| tm_min | [0.9, 0.9, 0.9] | vm_nom_fr/vm_nom_to | 3 values, one per phase |
| tm_max | [1.1, 1.1, 1.1] | vm_nom_fr/vm_nom_to | 3 values, one per phase |
| fixed | [true, true, true] | boolean | 3 values, one per phase |
| grounded_fr | true | boolean |  |
| grounded_to | true | boolean |  |

ccoffrin:
Ok one last thought about the data model.  I am not 100% sure this is the best way to go, but what do you think about making the value of `config_fr` a dict that includes all the information required about the connection?  For example, something like,
```
"config_fr":{
  "type":("wye"|"delta"|"z"),
  "cnd":[<Ints>],
  "grounded":<bool>
}
```
In this scenario maybe `"vm_nom"` should also be in the `config` block?

The advantage of this design is that it is well structured from a data standpoint.  The disadvantage is that it's a departure from the tabular format that is easier to read in a text file.  Without being close to the implementation it's not clear to me which is better.  What do you think?

sanderclaeys:
I like this idea. It groups together a set of properties that logically belong together. In this case I would opt for also adding vm_nom, as the config dictionary will contain all winding specific information. I added `polarity` as well, the equivalent of the plus/minus sign.

This data model is slightly more expressive than what is supported by the mathematical model. I added additional checks in `calc_tp_trans_Tvi` to ensure these additional properties are what we expect them to be.
- For now we only support grounded wye windings. This feature is not very interesting in Kron reduced networks.
- The secondary winding always has to be a 123+y winding (shown below in the new data model). Other options can be modelled as a composition of two transformers back-to-back. This is what is done automatically in the OpenDSS parser now. Long-term it might be desirable to support this directly, or map it from an intermediate level data model to a low level data model.
```
"config_to":{
  "type":"wye",
  "polarity":'+',
  "cnd":[1,2,3],
  "grounded":true
}
```

ccoffrin:
Sounds good to me!  @pseudocubic, I think I am good with this PR now.  Are you ready to merge this? 
 Do you want to tag any versions before we merge?

pseudocubic:
I'm good, I plan to tag after the julia v1.0 update (after this PR, and maybe the loads PR are merged, if we think that PR will be ready to merge before the end of next week)

ccoffrin:
I am confused.  So should we merge this now?  Or did you want to wait until dropping support for Julia v0.6/v0.7?

pseudocubic:
sorry, I typed that hastily and wasn't clear.

I decided to create a tag v0.1.5 that includes the minor units fix for the opendss parser. This will be the last release for 0.6. 

I am making a minor edit to disable 0.6 and 0.7 in this branch, and then I will merge this PR. 

If we think that PR #123 (the loads extension) can be merged before next Friday, I will wait to do my PR where I drop Julia < v1 support, otherwise I will do that first, and #123 will have to be refactored.

After these things have been completed I will create the last release before we do the MOI update.

ccoffrin:
Great, that is clearer now.

Let's not wait on sorting out #123 a forage ahead with this and dropping < v1.0 in what ever order makes sense.  It's not much work to make another release so if #123 resolves quickly no harm in making two releases in the span of a week or two.

PR #119, opened 2019-04-04
acr formulation

sanderclaeys:
This PR adds support for rectangular voltage coordinates, similar to the acr formulation in PMs. 

The default behaviour of PMs is to initialize all voltage phasors at (vr=1, vi=0). This leads to local infeasibility issues with Ipopt. To fix this, the starting point should be closer to the final solution, which tends to be in the vicinity of a balanced phasor. Rather than re-implementing the behaviour of variable_voltage, PMs.variable_voltage is called and then the starting point is overwritten in TPPMs.

Most of the unit tests for the acp formulation in test/tp_pf.jl and est/tp_opf.jl have been duplicated for acr, except for the matlab parser related ones. Nearly all of these worked out of the box, except for one case where the accuracy for 'vm' had to be reduced from 1E-4 to 1.5E-4.

This PR depends on [minor fixes in PMs](https://github.com/lanl-ansi/PowerModels.jl/pull/479), in two ways
- correction in PMs.constraint_voltage to index vmin and vmax properly
- fix for the solution builder, needed to run the new unit tests successfully


src/form/acr.jl
```diff
@@ -0,0 +1,178 @@
...
+            @constraint(pm.model, vi <= 0)
+        end
+    end
+end
+
+
+""
+function constraint_kcl_shunt_slack(pm::GenericPowerModel{T}, n::Int, c::Int, i::Int, bus_arcs, bus_arcs_dc, bus_gens, bus_pd, bus_qd, bus_gs, bus_bs) where T <: PMs.AbstractACRForm
+    vr = var(pm, n, c, :vr, i)
+    vr = var(pm, n, c, :vi, i)
```
pseudocubic: you have `vr` twice here, the second should be `vi`

sanderclaeys:
All issues for ACR have been adressed now; I also added a line to the changelog.

PR #123, opened 2019-04-17
load models extension

### LINKED ISSUE #10: ZIP Loads
Generalize the load model from constant power to ZIP.

sanderclaeys:
This PR extends the transformer extension, and should best be merged after. I decided to submit it already now to start the discussion on the load modelling.

This PR adds support for 
- constant current loads (proportional to vm)
- constant impedance loads (proportional to vm^2)
- wye and delta connected loads

These are all the load features needed to simulate the IEEE13 test feeder. General ZIP models are not included, but are easy to add based on the code added in this PR if desired. This PR therefore partially resolves #10.

The load models got their own problem definition, allowing easy comparison with the traditional constant power loads. Using the existing problem definitions, runs all other load models (constant current and constant impedance) as constant power loads.

The code complexity is due to two reasons:
1. OpenDSS allows very odd ‘connections’ of the loads, without throwing an error. This made the parser quite complex, and also required many unit tests to ensure that the parser behaviour is consistent with what TPPM does. I implemented all edge cases for one-phase loads, a subset for three-phase loads, and omitted two-phase loads (see discussion at the end).
2. The non-linear implementations of the delta loads are quite tedious. The load power at the bus requires 4 operations (division, multiplication, summation and subtraction), all of which have to be split in real components.

OpenDSS parser
I explicitly chose not to support two phase loads. Take the following example, which is legal in OpenDSS syntax:
`New Load.L1 nphases=2 bus1=x.1.2.0 conn=delta kw=… kvar=… kv=…`
This will connect a first load between phase 1 and 2, and a second load between phase 2 and the neutral. According to the internal model of TPPM, this would thus map to a combination of a single-phase delta load and single-phase wye load, although the load is specified to be delta (!).


ccoffrin:
Can you modify this PR to point to the transformer branch?  That way I can focus only on the load modeling extensions?

sanderclaeys:
I merged the updated transformer branch into this PR, and included a fix as discussed in #125.

As long as the transformer PR is not merged, there is no branch I can point to in TPPM which includes the transformer extension.

sanderclaeys:
Using this extension, I matched the results of OpenDSS on IEEE13 with 
- an accuracy of 1.3E-4 on the voltage magnitudes
- identical power drawn by the loads after rounding (up to the accuracy reported by OpenDSS, 100 Watt)

sanderclaeys:
I will submit a new PR that shows the diffs correctly

Issue #125, opened 2019-04-18
tp_pf voltage magnitude bound


sanderclaeys:
In the tp_pf problem formulation, the following constraint should be removed:
https://github.com/lanl-ansi/ThreePhasePowerModels.jl/blob/8abe80da4880382bc5fedaee2dab8e751bdd9733/src/prob/tp_pf.jl#L39
The purpose of this constraint is to enforce the voltage magnitude bounds in the ACR formulation as a SOC constraint. It is empty for the ACP formulation.

It should not be present in the power flow problem. I ran into this issue when debugging the load models.

ccoffrin:
I disagree.  Here is an example of a non-convex problem that requires it, https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/form/act.jl#L15.  Obscure, but worth further exploration.

ccoffrin:
If a specific formulation does not require it, just make an empty function like so, https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/form/acp.jl#L14

frederikgeth:
Ah, ok, I think we may be mixing up two things.

The purpose of `constraint_voltage` is to define KVL over all loops in a meshed network, and thereby reflect the part of the power flow physics that isn't captured by Ohm's law and the conservation of current.

However, for the ACR formulation, this constraint is actuallt used to apply bounds on the voltage magnitude, so voltage magnitude bounds are used, even when solving a power flow (even in PowerModels proper, I think). 

The use of `constraint_voltage` for this purpose seems consistent with the first use case.

ccoffrin:
Oh, I think we are having some confusion due to my lack of documentation.  

My intended semantics for `constraint_voltage` is for constraints that link multiple voltage variables.  For example the bounds in ACR or voltage-based inequalities in the relaxations.  If you wish to do KVL, then I would add new constraint (e.g. `constraint_kvl(...)`) which takes a collection of ordered bus ids which define a cycle in the network graph.

sanderclaeys:
I see. In that case, shouldn't there be a keyword argument 'bounded=false' which deactivates the bound setting for the ACR case? 
Because right now, for the power flow problem formulation, vm>0.9 is enforced for the ACR case, but not the ACP case.

ccoffrin:
Yes!  That is a bug in the current PF formulation.

ccoffrin:
We need an issue or PR on PowerModels to fix this.

[...]

Issue #122, opened 2019-04-15
OpenDSS phases=3 parser bug


sanderclaeys:
I encountered a bug for OpenDSS load specifications with `phases=3`.  The parsed  results are three times bigger than they should be. 

I am currently working on an ACP implementation of delta, wye, constant power, constant current and constant impedance loads. All of these are required for common IEEE unbalanced test cases. Related to #10.

I propose to include the fix for this directly in there, rather than in a separate PR. Is that okay for you guys?

For example, take
`new Load.L1 phases=3 kw=5`
This is parsed as a  load with
`pd=[5, 5, 5]`
Whilst the OpenDSS behaviour is
`pd=[5/3, 5/3, 5/3]`

pseudocubic:
Closed by #127

PR #127, opened 2019-04-27
load models extension

### LINKED ISSUE #10: ZIP Loads

sanderclaeys:
Note: same as #123, diffs are now showing correctly

This PR adds support for 
- constant current loads (proportional to vm)
- constant impedance loads (proportional to vm^2)
- wye and delta connected loads

These are all the load features needed to simulate the IEEE13 test feeder. General ZIP models are not included, but are easy to add based on the code added in this PR if desired. This PR therefore partially resolves #10.

The load models got their own problem definition, allowing easy comparison with the traditional constant power loads. Using the existing problem definitions, runs all other load models (constant current and constant impedance) as constant power loads.

The code complexity is due to two reasons:
1. OpenDSS allows very odd ‘connections’ of the loads, without throwing an error. This made the parser quite complex, and also required many unit tests to ensure that the parser behaviour is consistent with what TPPM does. I implemented all edge cases for one-phase loads, a subset for three-phase loads, and omitted two-phase loads (see discussion at the end).
2. The non-linear implementations of the delta loads are quite tedious. The load power at the bus requires 4 operations (division, multiplication, summation and subtraction), all of which have to be split in real components.

OpenDSS parser
I explicitly chose not to support two phase loads. Take the following example, which is legal in OpenDSS syntax:
`New Load.L1 nphases=2 bus1=x.1.2.0 conn=delta kw=… kvar=… kv=…`
This will connect a first load between phase 1 and 2, and a second load between phase 2 and the neutral. According to the internal model of TPPM, this would thus map to a combination of a single-phase delta load and single-phase wye load, although the load is specified to be delta (!).


sanderclaeys:
Instead of changing the starting point of the voltage phasors in a new constraint `set_tp_va_start_if_unset`, I overloaded `variable_tp_voltage` for the acp case, where I changed the starting point to a balanced phasor. This should be more consistent with past suggestions.

This had a minor influence on past unit tests; most notably, the objective of the unit test "3-bus balanced pv" dropped by 0.75% due to this change, a change bigger than the specified tolerance.

pseudocubic:
Please add tests for the `warn` and `error` statements

pseudocubic:
If ZIP loads really are a small lift to get into this PR I think we'd like to see them here

ccoffrin:
I am not sure I am a fan of ZIP loads.  Lets hold that thought until I have had a chance to review this feature in detail.

pseudocubic:
@ccoffrin From the PR text: 

"General ZIP models are not included, but are easy to add based on the code added in this PR if desired."

Therefore I would argue that it should be included now while we're working on loads, in order to cover the most broad categories of loads for support in one fell swoop, rather than leaving it until some undetermined time in the future. Plus it would close #10. 

If it were too much work, I could see leaving it, since it isn't an "in-demand" feature necessarily, but it seems better to add it while this code is fresh than to figure it out later.

src/io/opendss.jl
```diff
@@ -194,20 +194,132 @@ function dss2tppm_load!(tppm_data::Dict, dss_data::Dict, import_all::Bool)
...
             expected_kv = tppm_data["basekv"] / sqrt(tppm_data["conductors"])
             if !isapprox(kv, expected_kv; atol=expected_kv * 0.01)
                 Memento.warn(LOGGER, "Load has kv=$kv, not the expected kv=$(expected_kv). Results may not match OpenDSS")
             end
 
             loadDict["name"] = defaults["name"]
             loadDict["load_bus"] = find_bus(name, tppm_data)
-            loadDict["pd"] = PMs.MultiConductorVector(parse_array(defaults["kw"] / 1e3, nodes, nconductors))
-            loadDict["qd"] = PMs.MultiConductorVector(parse_array(defaults["kvar"] / 1e3, nodes, nconductors))
+
```
pseudocubic: I believe this now adds the follow fields to the load dict, correct me if I'm mistaken: `conn`, `model`, `vnom_kv`

src/io/dss_parse.jl
```diff
@@ -952,6 +952,22 @@ function parse_busname(busname::AbstractString)
...
+    parts = split(busname, '.'; limit=2)
+    ret = []
+    if length(parts)==2
+        conds_str = split(parts[2], '.')
+        ret = [parse(Int, i) for i in conds_str]
+    end
+    return ret
+end
+
+
```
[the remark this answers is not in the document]
sanderclaeys: I initially had to add this when I was still trying to support `phases=2`loads; for these, the order in which the phases are mentioned is important. The order is discarded by `parse_busname`

I will check whether under the current feature set (1 phase and subset of 3 phases), it is possible to drop it.

src/io/opendss.jl
```diff
@@ -194,20 +194,132 @@ function dss2tppm_load!(tppm_data::Dict, dss_data::Dict, import_all::Bool)
...
+                Memento.warn(LOGGER, "$load_name: load model 7 not supported. Treating as model 1.")
+                model = 1
+            elseif model == 8
+            # 8: ZIP
+                Memento.warn(LOGGER, "$load_name: load model 8 not supported. Treating as model 1.")
+                model = 1
+            end
+            # save adjusted model type to dict, human-readable
+            model_int2str = Dict(1=>"constant_power", 2=>"proportional_vmsqr", 5=>"proportional_vm")
+            loadDict["model"] = model_int2str[model]
```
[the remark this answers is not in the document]
sanderclaeys: I initially started out with `constant current`, but this easily leads to misunderstandings as to what it actually means (as it happened for me). What actually happens, is
P = (P0/|V0|).|V|
Q = (Q0/|V0|).|V|
This implies that the magnitude of the current remains constant, but not the real and imaginary components of the current. So 'constant current magnitude' would be more precise, or '(power is) proportional to voltage magnitude'

test/opendss.jl
```diff
@@ -54,10 +54,23 @@
         Memento.setlevel!(TESTLOG, "info")
 
         @test_throws(TESTLOG, ErrorException,
-                     TPPMs.parse_file("../test/data/opendss/test_simple3.dss"))
+                   TPPMs.parse_file("../test/data/opendss/loadparser_error_1.dss"))
```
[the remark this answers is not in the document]
sanderclaeys: Could we perhaps create a subfolder to deal with the clutter for now? 

What you propose would be helpful here and in other cases.  A method could be added which takes a template dss file, and inserts a specified string at a certain position. This could be used to create new dss files on the fly, and run them through the parser in the tests.

src/io/opendss.jl
```diff
@@ -194,20 +194,132 @@ function dss2tppm_load!(tppm_data::Dict, dss_data::Dict, import_all::Bool)
...
+                model = 1
+            elseif model == 7
+            # 7: Constant P and quadratic Q (i.e., fixed reactance)
+                Memento.warn(LOGGER, "$load_name: load model 7 not supported. Treating as model 1.")
+                model = 1
+            elseif model == 8
+            # 8: ZIP
+                Memento.warn(LOGGER, "$load_name: load model 8 not supported. Treating as model 1.")
+                model = 1
+            end
```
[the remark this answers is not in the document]
sanderclaeys: Yes, correct.

src/io/opendss.jl
sanderclaeys: Correct. I think @ccoffrin will prefer renaming `vnom_kv` to `vm_nom`, similar to the transformer data model.

sanderclaeys:
- I added an OPF version, `tp_opf_lm`
- I added `specification.md`, and included `tp_pf_lm`, `tp_opf_lm` and `tp_opf_oltc`

@pseudocubic I will address the parser comments in the next round of revision

I think the ZIP models are another good example of how an 'intermediate data model' is needed. Components in this intermediate data model can then be decomposed to a set of basic components, which are supported by TPPM.
- The ZIP loads can be decomposed in a set of 'constant power', 'constant current' and 'constant impedance' loads. This is a cleaner solution for cases where one of the ZIP parameters is zero for example, because then that component can simply be omitted.
- What is done now for the transformers in the OpenDSS parser, could be part of converting the intermediate data model to the decomposed data model.

@frederikgeth, @ccoffrin, what do you think?

ccoffrin:
I agree that ZIP loads make sense at  higher level, but not necessarily in the mathematical model.

docs/src/specifications.md
```diff
@@ -0,0 +1,207 @@
+# Problem Specifications
```
ccoffrin: In general we should try to avoid large blocks of Julia code in the docs.  It is too burdensome to keep the docs up to date with the actual implementation.  Providing doc strings in-line with the implementation is a more maintainable approach.

src/core/constraint.jl
```diff
@@ -24,3 +24,105 @@ function constraint_tp_storage_loss(pm::PMs.GenericPowerModel, n::Int, i, bus, r
...
+qd = cq.|vm|
+sd = cp.|vm| + j.cq.|vm|
+The same remark applies on delta/wye as for the fixed setpoint.
+"""
+function constraint_tp_load_power_prop_vm(pm::PMs.GenericPowerModel, load_id::Int; nw=pm.cnw)
+    load = PMs.ref(pm, nw, :load, load_id)
+
+    vnom_kv = load["vnom_kv"]
+    vbase_kv_LL = PMs.ref(pm, nw, :bus, load["load_bus"])["base_kv"]
+    vbase_kv_LN = vbase_kv_LL/sqrt(3)
```
ccoffrin: To be consistent with PowerModels conventions, this load model should be in p.u. before getting to this point.

src/core/variable.jl
```diff
@@ -1,7 +1,8 @@
 ""
-function variable_tp_voltage(pm::PMs.GenericPowerModel; kwargs...)
+
+function variable_tp_voltage(pm::PMs.GenericPowerModel; nw=pm.cnw, kwargs...)
     for c in PMs.conductor_ids(pm)
-        PMs.variable_voltage(pm, cnd=c; kwargs...)
+        PMs.variable_voltage(pm, cnd=c; nw=nw, kwargs...)
```
ccoffrin: Following conventions, nw should come before cnd in the arguments list.

src/core/variable.jl
```diff
@@ -204,3 +204,28 @@ function variable_tp_oltc_tap(pm::PMs.GenericPowerModel; nw::Int=pm.cnw, bounded
         end
     end
 end
+
+
+"Power drawn by the load at the bus to which it is connected"
+function variable_load_flow(pm::PMs.GenericPowerModel; nw::Int=pm.cnw, cnd::Int=pm.ccnd)
```
ccoffrin: Can we get by with `variable_load`?  not sure what is meant by `flow` here.  maybe `variable_load_power`?

src/core/variable.jl
```diff
@@ -204,3 +204,28 @@ function variable_tp_oltc_tap(pm::PMs.GenericPowerModel; nw::Int=pm.cnw, bounded
...
+    variable_active_load_flow(pm, nw, cnd)
+    variable_reactive_load_flow(pm, nw, cnd)
+end
+
+
+"Active power drawn by the load at the bus to which it is connected"
+function variable_active_load_flow(pm::PMs.GenericPowerModel, nw::Int, cnd::Int)
```
ccoffrin: Should these be bounded from 0 to given value?  E.g. see the implementation of dispatchable loads in PowerModels.

src/form/acp.jl
```diff
@@ -1,6 +1,25 @@
...
 
 
+""
+function variable_tp_voltage(pm::PMs.GenericPowerModel{T}; nw=pm.cnw, kwargs...) where T <: PMs.AbstractACPForm
+    for c in PMs.conductor_ids(pm)
+        PMs.variable_voltage(pm, cnd=c; nw=nw, kwargs...)
+    end
+    # This is needed for delta loads, where division occurs by the difference
+    # of voltage phasors. If the voltage phasors at one bus are initialized
+    # in the same point, this would lead to division by zero.
```
ccoffrin: Another solution to this that would use the PowerModels infrastructure would be to check if the field `va_start` exists in the data model.  If not, add the default starting vector that you propose.  

We can also considering making these standard angle offsets a standard across the data model.  Open an issue to look into that.

src/form/acp.jl
```diff
@@ -291,3 +328,197 @@ function constraint_tp_oltc_flow(pm::PMs.GenericPowerModel{T}, nw::Int, i::Int,
...
+
+
+"""
+For a delta load, sd = (s_ab, s_bc, s_ca), but we want to fix s = (s_a, s_b, s_c)
+s is a non-linear transform of v and sd, s=f(v,sd)
+s_a = v_a*conj(s_ab/(v_a-v_b) - s_ca/(v_c-v_a))
+s_b = v_b*conj(s_ab/(v_a-v_b) - s_ca/(v_c-v_a))
+s_c = v_c*conj(s_ab/(v_a-v_b) - s_ca/(v_c-v_a))
+"""
+function constraint_tp_load_power_setpoint_delta(pm::PMs.GenericPowerModel{T}, nw::Int, load_id::Int, load_bus_id::Int, pd::PMs.MultiConductorVector, qd::PMs.MultiConductorVector) where T <: PMs.AbstractACPForm
```
ccoffrin: This function should have an assertion checking the number of conductors/phases in the network model.

src/form/acp.jl
```diff
@@ -291,3 +328,197 @@ function constraint_tp_oltc_flow(pm::PMs.GenericPowerModel{T}, nw::Int, i::Int,
...
+    q_a_nlexp = q_x(vm_a, va_a, ire_ab, iim_ab, ire_ca, iim_ca)
+    q_b_nlexp = q_x(vm_b, va_b, ire_bc, iim_bc, ire_ab, iim_ab)
+    q_c_nlexp = q_x(vm_c, va_c, ire_ca, iim_ca, ire_bc, iim_bc)
+    JuMP.@NLconstraint(pm.model, p_a==p_a_nlexp)
+    JuMP.@NLconstraint(pm.model, p_b==p_b_nlexp)
+    JuMP.@NLconstraint(pm.model, p_c==p_c_nlexp)
+    JuMP.@NLconstraint(pm.model, q_a==q_a_nlexp)
+    JuMP.@NLconstraint(pm.model, q_b==q_b_nlexp)
+    JuMP.@NLconstraint(pm.model, q_c==q_c_nlexp)
+end
```
ccoffrin: Potentially this can be simplified from a modeling standpoint, but we can save that as an issue for later.

src/form/acp.jl
```diff
@@ -291,3 +328,197 @@ function constraint_tp_oltc_flow(pm::PMs.GenericPowerModel{T}, nw::Int, i::Int,
...
+    p_c_nlexp = p_x(vm_c, va_c, ire_ca, iim_ca, ire_bc, iim_bc)
+    q_a_nlexp = q_x(vm_a, va_a, ire_ab, iim_ab, ire_ca, iim_ca)
+    q_b_nlexp = q_x(vm_b, va_b, ire_bc, iim_bc, ire_ab, iim_ab)
+    q_c_nlexp = q_x(vm_c, va_c, ire_ca, iim_ca, ire_bc, iim_bc)
+    JuMP.@NLconstraint(pm.model, p_a==p_a_nlexp)
+    JuMP.@NLconstraint(pm.model, p_b==p_b_nlexp)
+    JuMP.@NLconstraint(pm.model, p_c==p_c_nlexp)
+    JuMP.@NLconstraint(pm.model, q_a==q_a_nlexp)
+    JuMP.@NLconstraint(pm.model, q_b==q_b_nlexp)
+    JuMP.@NLconstraint(pm.model, q_c==q_c_nlexp)
```
ccoffrin: Same as above.

src/form/acp.jl
```diff
@@ -291,3 +328,197 @@ function constraint_tp_oltc_flow(pm::PMs.GenericPowerModel{T}, nw::Int, i::Int,
...
```
ccoffrin: it looks like some code is replicated across these functions.  Consider adding some internal helper functions (i.e., starting with `_`) to share basic functionality across these constraints.

src/form/acr.jl
```diff
@@ -26,9 +26,11 @@ function variable_tp_voltage(pm::PMs.GenericPowerModel{T}; kwargs...) where T <:
...
 
 
-"delegate back to PowerModels"
-function constraint_tp_voltage(pm::PMs.GenericPowerModel{T}, n::Int, c::Int) where T <: PMs.AbstractACRForm
-    PMs.constraint_voltage(pm, n, c)
+"only delegate back to PowerModels if bounded"
+function constraint_tp_voltage(pm::PMs.GenericPowerModel{T}, n::Int, c::Int, bounded::Bool) where T <: PMs.AbstractACRForm
+    if bounded
+        PMs.constraint_voltage(pm, n, c)
+    end
```
ccoffrin: This seems odd, shouldn't the "bounded" parameter be passed into the `constraint_voltage` voltage function?  Then PowerModels should skip the constraint?  It may require a bug fix in PowerModels, but that should not be addressed here.

src/form/acr.jl
```diff
@@ -145,3 +167,192 @@ q[t_idx] == -(b+b_to)*v[t_bus]^2 - (-b*tr+g*ti)/tm*(v[t_bus]*v[f_bus]*cos(t[f_bu
...
+end
+
+
+function constraint_load_power_prop_vmsqr_wye(pm::PMs.GenericPowerModel{T}, nw::Int, cnd::Int, load_id::Int, load_bus_id::Int, cp::Real, cq::Real) where T <: PMs.AbstractACRForm
+    pd = PMs.var(pm, nw, cnd, :pd, load_id)
+    qd = PMs.var(pm, nw, cnd, :qd, load_id)
+    vr = PMs.var(pm, nw, cnd, :vr, load_bus_id)
+    vi = PMs.var(pm, nw, cnd, :vi, load_bus_id)
+    JuMP.@NLconstraint(pm.model, pd==cp*(vr^2+vi^2))
+    JuMP.@NLconstraint(pm.model, qd==cq*(vr^2+vi^2))
```
ccoffrin: This looks like a bus shunt, or am I missing something?

src/form/acr.jl
sanderclaeys: Yes, ideally this should be addressed at PM level and not TPPM. This is a workaround for #125

src/form/acp.jl
sanderclaeys: So in this case,
- `p_a` is the load power variable
- `p_a_nlexp` is a NL expression

I can avoid creating explicit load power variables, and store `p_a_nlexp` directly in the key `p_a` in the variable Dict. Would you prefer that?

src/form/acp.jl
sanderclaeys: Comment above would remove these lines.

src/form/acp.jl
sanderclaeys: Comment above would remove these lines in all constraints.

src/form/acr.jl
sanderclaeys: Yes, the 'proportional_vmsqr' or 'constant impedance' loads are equivalent to bus shunts.

src/core/variable.jl
sanderclaeys: Agreed, let's go with `variable_load`

src/core/variable.jl
sanderclaeys: [This](https://github.com/lanl-ansi/PowerModels.jl/blob/f5289fcee04b4eba0f2a0221c729fb3efb5f74e5/src/core/variable.jl#L286) is the relevant line in the dispatchable branch I think. Why is the upperbound fixed to 1.0?

I can infer bounds via the nominal power drawn, and the voltage magnitude bounds for the bus to which the load is connected. So for example for the constant current load
P = (P0/V0).|V|, so
if P>=0
    P \in [(P0/V0).vmin, (P0/V0).vmax]
else
    P \in [(P0/V0).vmax, (P0/V0).vmin]

src/form/acp.jl
ccoffrin: That was the change I was thinking about, but it can have impacts on the NLP solver, so let's open an issue and revisit it later.

ccoffrin:
@sanderclaeys, can you tell me about what features are added to the data model to support this?

sanderclaeys:
Three new fields are added
- **`conn`** \in {`"wye",` `"delta"`}
- **`model`** \in {`"constant_power"`, `"proportional_vm"`, `"proportional_vmsqr"`} (names subject to change, as per above discussion)
- **`vm_nom`**

All loads still have fields 
- **`pd`**
- **`qd`**

Their meaning changes depending on model and connection. 

**`conn=wye`, `model="constant_power"`**
pd_wye_a = **`pd[1]`**

**`conn=wye`, `model="proportional_vm"`**
pd_wye_a = (**`pd[1]`**/**`vm_nom`**).vm_a

**`conn=delta`, `model="constant_power"`**
pd_wye_a  = v_a.conj(i_ab-i_ca) 
                  = v_a.[s_ab/(v_a-v_b) + s_ca/(v_c-v_a)]
                  = v_a.[(**`pd[1]`**+im.**`qd[1]`**)/(v_a-v_b) + (**`pd[3]`**+im.**`qd[3]`**)/(v_c-v_a)]

The benefit of keeping **`pd`** and **`qd`** unchanged, is that it remains possible to use old problem formulations on this problem.

sanderclaeys:
I finally updated the PR to the new master and JuMP 0.19.

I ran into a very weird issue however. When the model contains delta loads, the starting point of the voltage phasors cannot be the same, or there will be a division by zero. This is why we added a balanced initialization in the past.

After the update to JuMP 0.19 however, the model now seemingly randomly terminates with 'INVALID_MODEL', and otherwise with 'LOCALLY_SOLVED' when using Ipopt.jl. Using ipopt through AmplNLWriter.jl on the other hand, always returns 'LOCALLY_SOLVED'. I included a [script](https://github.com/lanl-ansi/ThreePhasePowerModels.jl/files/3230767/debug_ipopt.jl.txt) which demonstrates this difference.

For now I have added AmplNLWriter to the unit test for load models. This allows me to finish updating this PR with working unit tests.

sanderclaeys:
Something weird is happening with the tests; I will see tomorrow whether I can reproduce it locally.

sanderclaeys:
Okay, the loadmodels are working now with the ampl workaround. The remaining errors are probably due to parser changes, will fix those tomorrow.

sanderclaeys:
I changed the voltage base procedure to start at the sourcebus and not the primary of the first transformer. This will parse IEEE34 correctly by default. 

This affects those cases that do not have a consistent voltage base definition. Because of this, I had to change some unit tests accordingly. 

Next I will implement the changes we discussed in the call.

src/io/dss_parse.jl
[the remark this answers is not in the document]
sanderclaeys: I removed a few edge cases that we can only support when the phase order is known. These edge cases will now cause an error instead.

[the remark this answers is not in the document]
sanderclaeys: I also added a unit test to ensure this. The unsupported specifications are now in case3_lm_1230_old.dss

src/io/opendss.jl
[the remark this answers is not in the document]
sanderclaeys: Per discussion I went for the human readable ones mentioned above. This should make the code more readable than using the equivalent exponents of the exponential model, as beautiful as it is though

test/opendss.jl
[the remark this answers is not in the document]
sanderclaeys: I rewrote the tests as loops instead, as you suggested.

src/io/opendss.jl
```diff
@@ -194,20 +194,132 @@ function dss2tppm_load!(tppm_data::Dict, dss_data::Dict, import_all::Bool)
...
+                    # this is a wye load in the TPPM sense
+                    loadDict["conn"] = "wye"
+                    ph = (cnds[2]==0) ? cnds[1] : cnds[2]
+                    pqd_premul = zeros(3)
+                    pqd_premul[ph] = 1
+                else
+                    # this is a delta load in the TPPM sense
+                    loadDict["conn"] = "delta"
+                    pqd_premul = zeros(3)
+                    pqd_premul[delta_map[cnds]] = 1
```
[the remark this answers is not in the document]
sanderclaeys: Now that the order is no longer considered, I could simplify this manually, as only [1,2], [2,3] and [1,3] can still occur.

src/form/acp.jl
sanderclaeys: This is now simplified as these NLexpressions are saved directly in the variable dict.

src/form/acp.jl
sanderclaeys: This is now implemented.

src/core/constraint.jl
sanderclaeys: If the vbase of the load is not saved, it has to be guaranteed that the bus to which the load is connected, and the load, have the same vbase. This is hard to do as long as the p.u. handling is spread out accross opendss.jl for the TPPM components, and delegated to PM for all PM components. For now, I propose to keep this here until p.u. is dealt with explicitly in one place for TPPMs.

src/form/acp.jl
sanderclaeys: I added this at the template level in constraint.jl

ccoffrin:
Let's make sure to get a working example of the Ipopt issue posted as an issue in JuMP.jl / Ipopt.jl.

The current AMPLNLWriter work around is not ideal (although I will accept it in the near term), so we should help the JuMP devs resolve the inconsistency.

sanderclaeys:
I created a working example in [this branch](https://github.com/sanderclaeys/ThreePhasePowerModels.jl/tree/snippet). The working example is in a script in the root directory, 'ipopt_debug.jl', and relies on the code in that branch to build the model. I tried as much as possible to strip away PowerModels specific functions and use pure JuMP

sanderclaeys:
With the bug fix in JuMP, the workaround is no longer needed. I removed it and updated the JuMP dependency in the Project.toml file.

pseudocubic:
I'd like to merge this in the near future so that I can move on to the PowerModels v0.12 upgrade, @sanderclaeys can you confirm that you have made the changes we all discussed over the phone and that this PR is ready? Please also ensure that the travis tests are passing, I see that they are not at the moment. Once you are ready @ccoffrin and I can review one last time.

sanderclaeys:
I have made the changes we discussed, the PR is ready for review. 

The tests pass on my laptop, the issue is caused by the JuMP 0.19.2 requirement. I changed the requirement in `Project.toml` to `^0.19.2`. Travis errors with
`ERROR: empty intersection between JuMP@0.19.1 and project compatibility 0.19.2-0.19`
@pseudocubic do you know how to solve this for Travis?

pseudocubic:
@sanderclaeys it was failing because when you updated the `compat` section of Project.toml, you didn't also do a `]up` to auto generate the new Manifest.toml.

pseudocubic:
@ccoffrin

sanderclaeys:
@pseudocubic can we prioritize this PR? 

My work the next weeks relies on extending this PR, so it would be nice to have this in the master. Also, this solves #122, which affects the resulting voltage magnitudes quite dramatically (I rediscovered this today the hard way)

ccoffrin:
Sure thing, I will do a review tomorrow.

src/prob/tp_pf_lm.jl
```diff
@@ -0,0 +1,101 @@
...
+""
+function run_tp_pf_lm(file::String, model_constructor, solver; kwargs...)
+    data = ThreePhasePowerModels.parse_file(file)
+    return PMs.run_generic_model(data, model_constructor, solver, post_tp_pf; multiconductor=true, kwargs...)
+end
+
+
+""
+function post_tp_pf_lm(pm::PMs.GenericPowerModel)
+    add_arcs_trans!(pm)
```
ccoffrin: Add a maintenance issue, `add_arcs_trans!`, should be added as `ref_extention` rather than being called in post_*.  In fact we may consider a more TPPM extension of run_* that includes this.

src/core/variable.jl
```diff
@@ -1,7 +1,8 @@
 ""
-function variable_tp_voltage(pm::PMs.GenericPowerModel; kwargs...)
+
+function variable_tp_voltage(pm::PMs.GenericPowerModel; nw=pm.cnw, kwargs...)
     for c in PMs.conductor_ids(pm)
-        PMs.variable_voltage(pm, cnd=c; kwargs...)
+        PMs.variable_voltage(pm, nw=nw, cnd=c, kwargs...)
```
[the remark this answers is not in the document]
sanderclaeys: I was not sure that further down nw is set to pm.cnw by default, but indeed obsolete to do it here

src/core/constraint_template.jl
```diff
@@ -264,3 +270,40 @@ function constraint_kcl_shunt_trans(pm::PMs.GenericPowerModel, i::Int; nw::Int=p
...
+    model = PMs.ref(pm, nw, :load, id, "model")
+    if model=="constant_power"
+        constraint_tp_load_power(pm, id)
+    elseif model=="constant_current"
+        constraint_tp_load_current(pm, id)
+    elseif model=="constant_impedance"
+        constraint_tp_load_impedance(pm, id)
+    else
+        Memento.@error(LOGGER, "Unknown model $model for load $id.")
+    end
```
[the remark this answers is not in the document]
sanderclaeys: I changed it to a single level of indirection already. I merged the docstrings into a lengthy one in constraint_template

sanderclaeys:
I think that should settle it. 

@ccoffrin For future reference, let me just make sure I get what you intend for the docstrings. Right now, `specifications.md` contains an exact copy of the code of each the problem formulation. In order to move this to the docstrings, the docstring of each post method should contain a full copy of the entire post method definition. Is this what you have in mind?

PR #129, opened 2019-04-29
voltage unbalance constraints

sanderclaeys:
This PR adds support for voltage unbalance constraints. This is a common requirement in distribution networks. For example, IEC standard 61000-2-2 and EN5160 requires that the voltage unbalance factor (VUF) should be below 2%.

Voltage unbalance constraints are defined on (ratios of) the sequence component magnitudes. A voltage phasor can be transformed to symmetric components: a positive, negative and zero sequence component. The 'voltage unbalance factor' (VUF) is the ratio of the magnitude of the negative sequence component to the magnitude of the positive sequence component.

A great discussion of this topic can be found below. We only implemented the VUF constraint, because EN5160 applies in Europe to residential customers and is of more interest to us.
[Kshitij Girigoudar, Daniel K. Molzahn, Line A. Roald, _Analytical and Empirical Comparisons ofVoltage Unbalance Definitions_](https://molzahn.github.io/pubs/girigoudar_molzahn_roald-2019.pdf)

A new problem definition was added, tp_opf_ubctr. In the data model, the following additional properties can be defined on buses:
- **vufmax**: voltage unbalance factor upper limit
- **vmposmax**: magnitude of positive sequence upper limit
- **vmnegmax**: magnitude of negative sequence upper limit
- **vmzeromax**: magnitude of zero sequence upper limit

If one of these properties is present, an additional constraint is added for that bus. Also, an additional solution_builder was defined that calculates the sequence components of the phasors.

As a test case, we defined a problem where the unbalance constraint (all of the above are tested) should be binding. In the unit test, we check that this constraint is indeed binding, and that the voltage phasor has the expected sequence components and VUF.


ccoffrin:
I propose to have a call next week to discuss these changes.  I am generally onboard with the idea but I would like to bikeshed on naming conventions.

I would like to try to void introducing the concepts of pos/neg/zero sequence, if possible.

sanderclaeys:
@frederikgeth also raised an issue with three-wire segments. In a three-wire segment (between the delta side of two transformers for example), the difference of the phase voltages should be bounded, and not their absolute value.

This could be a good moment to also add those voltage constraints.

sanderclaeys:
In summary, a single function is added in the problem formulation, `constraint_tp_balance`, which applies the other constraints if the corresponding key is present in the data model of the bus.
- `vm_pos_seq_max`, `vm_pos_neg_max` and `vm_pos_zero_max`
- `vuf_max`
- `vm_ll_max`, `vm_ll_min`

My goal was to implement the discussed structure to continue the discussion. Once we converge on this, I will add the ACR version and add more unit tests. I will need to make a reference case manually, as this is inherently OPF and not PF.

pseudocubic:
I'd like to merge this in the near future so that I can move on to the PowerModels v0.12 upgrade, @sanderclaeys are you waiting for additional feedback, or is this ready for a review and merge?

sanderclaeys:
This is ready for review. I will add some more unit tests next week to cover all constraint types, these features are not part of OpenDSS so validating them is less straightforward.

pseudocubic:
@ccoffrin

sanderclaeys:
@pseudocubic I added some more unit tests, it is all ready now

pseudocubic:
@ccoffrin this is up to date with the new master and looks ready to merge to me, any additional comments?

src/core/constraint_template.jl
```diff
@@ -272,6 +272,36 @@ function constraint_kcl_shunt_trans(pm::PMs.GenericPowerModel, i::Int; nw::Int=p
 end
 
 
+"Impose all balance related constraints for which key present in data model of bus."
+function constraint_tp_voltage_balance(pm::PMs.GenericPowerModel, bus_id::Int; nw=pm.cnw)
```
ccoffrin: Should we have a bibtex ref to a related paper in this doc string?

src/core/constraint_template.jl
```diff
@@ -272,6 +272,36 @@ function constraint_kcl_shunt_trans(pm::PMs.GenericPowerModel, i::Int; nw::Int=p
...
+
+    if haskey(bus, "vm_zero_seq_max")
+        constraint_tp_vm_zero_seq(pm, nw, bus_id, bus["vm_zero_seq_max"])
+    end
+
+    if haskey(bus, "vm_ll_min")|| haskey(bus, "vm_ll_max")
+        vm_ll_min = haskey(bus, "vm_ll_min") ? bus["vm_ll_min"] : PMs.MultiConductorVector(fill(0, 3))
+        vm_ll_max = haskey(bus, "vm_ll_max") ? bus["vm_ll_max"] : PMs.MultiConductorVector(fill(Inf, 3))
+        constraint_tp_vm_ll(pm, nw, bus_id, vm_ll_min, vm_ll_max)
+    end
```
ccoffrin: Post an issue to add some overview documentation for these different types of balance constraints.

src/core/solution.jl
```diff
@@ -0,0 +1,30 @@
...
+    if !PMs.ismultinetwork(pm)
+        get_solution_vm_all(pm, pm.cnw, sol)
+    else
+        for nw in PMs.nws(pm)
+            get_solution_vm_all(pm, nw, sol[nw])
+        end
+    end
+end
+
+function get_solution_vm_all(pm::PMs.GenericPowerModel, nw::Int, sol_nw::Dict{String,<:Any})
```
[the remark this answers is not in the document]
sanderclaeys: I had to use the function names of before PowerModels 0.12, because TPPM is still at 0.10 and this would break the unit tests in the meanwhile.

ccoffrin:
LGTM, @pseudocubic merge at your convenience.

PR #138, opened 2019-05-17
capacitor bug fix

sanderclaeys:
Currently there are two bugs in the parsing of capacitors
- The base should be adjusted as before
- The sign is wrong. The power reference in OpenDSS is expressed in generator sign convention, but the parser  assumes load convention. This can be seen by observing that a capacitor creates a voltage drop instead of a voltage rise.

A unit test will follow later, as part of the matrix line charging parameters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ ThreePhasePowerModels.jl Change Log
 ### staged
 - Allow for arbitrarily named sourcebus
 - Add json parser
+- Fix bug in OpenDSS parse of Capacitors [zbase factor and wrong sign] (#138)
 
 ### v0.3.0
 - Update to JuMP v0.19/MathOptInterface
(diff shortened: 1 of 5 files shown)


pseudocubic:
I would prefer there to be a unit test for this, perhaps by adding a simple capacitor to a 3 bus case?

A note also needs to be added to the change log

sanderclaeys:
I will get around to it this week. I will skip the matrix line shunts for now; it is easy enough to extended the ACP/ACR formulations but the BF ones will require more work

sanderclaeys:
I added a unit test for the capacitor pf. A few values had to be updated in other tests (4 in total I think)

ccoffrin:
LGTM.  

@pseudocubic, regarding the broken docs build, most likely the manifest.toml needs to be removed form the docs dir.

PR #140, opened 2019-05-17
ADD: JSON parser

pseudocubic:
Adds json parsing capability to `parse_file` by relying on PowerModels json parser, but using TPPMs validity checks.

Adds unit tests and updates changelog


Project.toml
```diff
@@ -17,6 +17,7 @@ Cbc = ">= 0.4"
 InfrastructureModels = "^0.2"
 Ipopt = ">= 0.4"
 JuMP = "^0.19.1"
+JSON = ">= 0.20"
```
ccoffrin: I think this bound can be v0.18.  I think that is when they added julia v1.0 support.

PR #141, opened 2019-06-03
UPD: PowerModels v0.11

pseudocubic:
Updates for PowerModels v0.11, and matches syntax changes introduced in new version of PMs:

- Functions starting with `_` indicates internal functions, not useful for users
- `correct_` indicates function will change data
- `check_` indicates function will only throw warnings / errors, not alter data
- Functions ending in `!` will alter data, not have a return

This is a breaking change, and will be included in TPPMs v0.4


ccoffrin:
I did not check the PR in detail, but the overview sounds good to me!

If you have updated to the `_` internal function naming convention, did you also add the JuMP template for exporting functions?  For example, see the comment out text here, https://github.com/lanl-ansi/PowerModels.jl/blob/master/src/core/export.jl

This will be the export convention for PowerModels v0.12

PR #151, opened 2019-07-09
ADD: load shedding problem (MLD)

pseudocubic:
Adds support for the load shed problem for the `ACPPowerModel`, `NFAPowerModel` and `LPLinUBFPowerModel`. Storage is supported for the ACP and NFA formulations, but not the LinDistFlow model. Adds these new problem definitions:

- `run_tp_mld`
- `run_tp_mld_strg`
- `run_tp_mld_uc`
- `run_tp_mld_bf`

Adds minimum load delta objectives:

- `objective_tp_min_load_delta`
- `objective_tp_min_load_delta_strg`
- `objective_tp_max_loadability` # example only, not used

Adds variables:

- `z_demand`
- `z_shunt`
- `z_gen`
- `z_voltage`
- `z_storage`

Adds unit tests, documentation, and updates change log.


src/io/opendss.jl
```diff
@@ -1051,9 +1051,9 @@ function _dss2pmd_storage!(pmd_data::Dict, dss_data::Dict, import_all::Bool)
             storageDict["discharge_rating"] = defaults["%discharge"] * defaults["kwrated"] / 1e3 / 100.0
             storageDict["charge_efficiency"] = defaults["%effcharge"] / 100.0
             storageDict["discharge_efficiency"] = defaults["%effdischarge"] / 100.0
-            storageDict["thermal_rating"] = _PMs.MultiConductorVector(_parse_array(defaults["kva"] / 1e3, nodes, nconductors))
-            storageDict["qmin"] = _PMs.MultiConductorVector(_parse_array(-defaults["kvar"] / 1e3, nodes, nconductors))
-            storageDict["qmax"] = _PMs.MultiConductorVector(_parse_array( defaults["kvar"] / 1e3, nodes, nconductors))
+            storageDict["thermal_rating"] = _PMs.MultiConductorVector(_parse_array(defaults["kva"] / 1e3 / nconductors, nodes, nconductors))
+            storageDict["qmin"] = _PMs.MultiConductorVector(_parse_array(-defaults["kvar"] / 1e3 / nconductors, nodes, nconductors))
+            storageDict["qmax"] = _PMs.MultiConductorVector(_parse_array( defaults["kvar"] / 1e3 / nconductors, nodes, nconductors))
```
ccoffrin: @frederikgeth or @sanderclaeys can you double check this interpretation of these ratings?

frederikgeth: This seems fine, we'll need to detail this in the eventual paper. We essentially have to change the semantics of the OpenDSS file, as kvar normally defines a set point, not bounds. But if you don't do this, there is nothing to optimize...

ccoffrin:
We should discuss the pros/cons of doing this, `_PMs.var(pm, nw, cn)[:z_shunt] = _PMs.var(pm, nw, :z_shunt)` as compared to having the constraints lookup the `z` variable at `_PMs.var(pm, nw, :z_shunt)`.

pseudocubic:
@ccoffrin that was left over code and wasn't needed, I already had moved almost all of the code for indicator variables to be at `var(pm, nw, :z_...)`, and just needed to do a little more cleanup.

Issue #152, opened 2019-07-30
Function naming scheme: `_tp_`


pseudocubic:
Currently, all PowerModelsDistribution variables/constraints/etc. start with e.g. `constraint_tp_`, but with the package rename and move to using the `conductor` name convention in PowerModels, I am not sure `tp` continues to make sense, especially for the run functions with which most users will actually interact. 

I am inclined to use `mc` for multiconductor, but am open to suggestions, discussion.

@ccoffrin @frederikgeth @sanderclaeys

sanderclaeys:
I don't feel strong about the specific name; mc seems like a good alternative to tp.

So going forward, the convention is that any constraint that does not take the conductor id as an argument, should start with constraint_mc (or alternative to mc)?

ccoffrin:
I agree it is good to revisit these constraint names given the package name update.  Given our discussion last week I think `mc` is a good indicator for modeling layer, but the constraint template will likely have to dispatch to more specific constraints like `3p` and `3pn` for three phase and three phase and a neutral.

We would need to get into the details to understand the best software design, but a simple replacement of `_tp_` to `_mc_` seems like a good quick fix for now.

sanderclaeys:
I think it might be a good idea as well to use something like
- **mp** = multi-phase, for constraints that have no (explicit) neutral and multiple phases
- **3p** three-phase, when a constraint is hard-coded for three phases
- **mpn/mc** multi-conductor, with an explicit neutral (which is the last conductor by convention?)
- **3pn**

For now I will use **mc** everywhere.

MartaVanin:
Hello,
first of all, thank you a all lot for the very good work you are doing!
I just have a comment wrt this thread: currently in PowerModelsDistribution, the ohms constraints still require a loop over the conductors, even though they contain `_tp_` in their name, e.g. constraint_ohms_tp_yt_from(pm::GenericPowerModel, i::Int; nw::Int=pm.cnw, cnd::Int=pm.ccnd).
I understand it is because they need it to account for mutual coupling between different conductors later, but then I find it a bit confusing as the user needs to add the loop explicitly, while the name kind of suggests that it shouldn't be needed.

pseudocubic:
@MartaVanin this is definitely a good point, and ultimately we hope to correct this issue, see #81 and #61

PR #153, opened 2019-08-14
matrix line shunts

sanderclaeys:
This PR contains most of the changes needed to implement matrix branch shunts. Open issues are
- `constraint_tp_ohms_yt_to_on_off` has not been extended yet. It does not appear in the unit tests in a breaking way and I am not sure how/where it is used, so I did not extend it.
- The constraint on the voltage angle difference is handled at PM level, `_PMs.constraint_voltage_angle_difference`. This method uses the branch shunts, which are considered to be vectors at PM level. This does not throw an error, because e.g. branch["b_to"][2] will return element (1,2) without throwing an error. 
@ccoffrin @frederikgeth @pseudocubic I normally never use this constraint, so I first wanted to discuss this. Should we completely reimplement this at PMD level or do  you see another solution?

The PR does not contain unit testing with branch shunts with non-zero off diagonals. I will add this in the next days and validate it against OpenDSS.


ccoffrin:
Regarding `constraint_tp_ohms_yt_to_on_off`, yes I think this can be ignored (deleted even, @pseudocubic?).  At one point were going to branch-based topology optimization, but now the switch-based method are the primary preference.

Regarding `constraint_voltage_angle_difference` this the implementation for ACP and ACR in PowerModels should be correct.  For the WR forms as long as it looks up on the diagonal of the W matrix it should also be correct.  For Branch Flow probably some significant derivation is required; if so, add an issue and make this constraint do nothing in the multi-phase Branch Flow models for now.

src/form/bf_mx.jl
```diff
@@ -95,8 +95,8 @@ function variable_tp_branch_series_current_prod_hermitian(pm::_PMs.GenericPowerM
         y_to_mag = abs.(branch["g_to"].values + im* branch["b_to"].values)
 
         smax = branch["rate_a"].values
-        cmaxfr = smax./vmin_fr + vmax_fr.*y_fr_mag
-        cmaxto = smax./vmin_to + vmax_to.*y_to_mag
+        cmaxfr = smax./vmin_fr + y_fr_mag*vmax_fr
+        cmaxto = smax./vmin_to + y_to_mag*vmax_to
```
ccoffrin: It's not immediately clear to me that this is correct.  Did you do a derivation?

sanderclaeys: Yes, I looked at these valid bounds in detail when I worked on the bf rewrite. The current is given by
`I = Y.V + conj(S./V)`

This implies a valid bound on the magnitude of I, |I|, as a function of Smax, Vmax and Vmin
`|I| <= |Y|.|V| + |S|./|V| <= |Y|.Vmax + Smax./Vmin`

In scalar form it is clearer that this is correct. 
`I[c] = sum(Y[c,d]V[d] over d) + conj(S[c]./V[c])`

The valid bound is arrived at by repeatedly applying the triangle inequality.
`|I[c]| = |sum(Y[c,d]V[d] over d) + conj(S[c]./V[c])| <= |sum(Y[c,d]V[d] over d)| + |conj(S[c]./V[c])|  <= sum(|Y[c,d]|.|V[d]| over d) + |S[c]|./|V[c]|`

sanderclaeys: Perhaps it is possible to derive a tighter valid bound by rewriting the equality first, but at least this is a valid extension to the matrix branch shunts.

sanderclaeys:
For the function `constraint_voltage_angle_difference`, only the `AbstractBFForms` implementation depends on the branch shunts. I duplicated this method at PMD level, and updated the indexing to work with matrix branch shunts, ignoring the off-diagonals. 

All old unit tests work now. In the future, this method should be extended to consider the off-diagonals as well. I will create an issue for this.

src/form/bf_mx.jl
ccoffrin: valid is sufficient, thanks for the explanation.

pseudocubic:
Yes, probably the on_off constraint could be removed, definitely you may ignore it for the purposes of this PR; I will revisit this constraint during function renaming discussed in #152

src/core/ref.jl
```diff
@@ -136,3 +136,14 @@ function _calc_tp_trans_Tvi(pm::_PMs.GenericPowerModel, i::Int; nw=pm.cnw)
     Cv_to *= vmult
     return (Tv_fr,Tv_im,Ti_fr,Ti_im,Cv_to)
 end
+
+
+function make_multiconductor!(mp_data, n_conductors::Int)
```
pseudocubic: I think this function should go into `src/core/data.jl`

ccoffrin: correct!

ccoffrin: good catch

sanderclaeys:
@frederikgeth @ccoffrin 

Is there any documentation/paper on how the existing PAD are derived for the multi-phase BFM, without matrix shunts? 

The current implementation only seems to use the diagonal elements of the line impedance (e.g. `r = branch["br_r"][c,c]`), so I am unsure whether this also correct when the the phases are linked (have non-zero off-diagonals).

I see a way to derive some valid bounds, but I do not see a way to represent the PAD exactly in multi-phase lifted BFM.

ccoffrin:
I think extending the PAD/VAD constraints to multi-phase is a point that requires some careful consideration that has not been considered thus far.  For the polar form phase-by-phase should be the same as PowerModels, for Branch Flow it is not as clear.  In any case, I do think this is a low priority constraint as we are the only tool that would support it.

As for the derivation, have a look at Section 5 of [this paper](https://arxiv.org/abs/1506.04773).

sanderclaeys:
@ccoffrin Thanks, that was very helpful. I derived the [equivalent for multi-phase systems](https://github.com/lanl-ansi/PowerModelsDistribution.jl/files/3516126/mpvad.pdf). 

@frederikgeth can you review this to make sure it is correct? It would be nice if we can include this in some paper, perhaps the real-value equivalent PSCC one. Then we can refer to a published derivation in the future.

If we agree on this I will add this and resolve it together with the other comments.

ccoffrin:
Nice Sander!  From a quick scan the derivation looks correct to me.

frederikgeth:
Eq 14 is correct indeed. (see draft of the "Nonconvex Lifted Unbalanced Branch Flow Model" paper eq24 for the same expression)

sanderclaeys:
Some BFM linearizations do not have full matrix variables but only the diagonals. In these cases, I cannot use the derivation above, which needs the full matrix variables.

Therefore, instead, I propose to 'patch' the BFM formulations so they can handle matrix shunts in the data model but only use the diagonal as before. Then once we merged #150, I will update BFM also conceptually to reflect the off-diagonals of the shunts. Doing major changes to BFM before #150 is merged, will lead to more issues when merging #150.

frederikgeth:
I have no problem with the approach you propose, as long as we open an issue to not forget about it.

src/core/constraint_template.jl
```diff
@@ -498,3 +498,18 @@ function constraint_tp_voltage_magnitude_sqr_on_off(pm::_PMs.GenericPowerModel,
 
     constraint_tp_voltage_magnitude_sqr_on_off(pm, nw, cnd, i, bus["vmin"][cnd], bus["vmax"][cnd])
 end
+
+
+"This is duplicated at PMD level to correctly handle the indexing of the shunts."
+function constraint_voltage_angle_difference(pm::_PMs.GenericPowerModel, i::Int; nw::Int=pm.cnw, cnd::Int=pm.ccnd)
```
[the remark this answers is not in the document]
sanderclaeys: I went with `mc` in the problem definition, and `mls` for the patched version of BFM. Later on we can then remove this and directly implement the `mc` version.

sanderclaeys:
I will add a unit test tomorrow, and then it should be ready for merging.

sanderclaeys:
I added a new test case with large matrix branch shunts, `case_mxshunts.dss`. Both the ACR and ACP formulation match up to OpenDSS with an accuracy of 1E-8.

Setting the off-diagonal elements to zero, affects the voltage magnitude on the order of 1E-3. Therefore, I am very confident that the off-diagonals have been modeled correctly in this PR.

PR #158, opened 2019-08-21
Parser fix

sanderclaeys:
This PR contains two minor changes
- There was a bug in the capacitor parsing, where the base voltage was used instead of the specified nominal voltage.
- The voltage base parsing starts at the sourcebus instead of the primary of the first transformer; this gives a nicer result for IEEE34.


sanderclaeys:
Additionally, single-phase capacitors were not parsed correctly. I ended up rewriting the relevant code to make it more understandable. I will update the broken unit tests (case5_phase_drop.dss contained a single-phase capacitor) to reflect the corrected parsing.

sanderclaeys:
I had to update the test case as well; the linecode specifications would lead to different behaviour in OpenDSS than expected. In its current format, PMD yields the same results as OpenDSS.

src/io/opendss.jl
```diff
@@ -355,17 +355,27 @@ function _dss2pmd_shunt!(pmd_data::Dict, dss_data::Dict, import_all::Bool)
...
+            qnom = (defaults["kvar"]/1E3)/defaults["phases"]
+            b_cap = qnom/vnom_ln^2
+            #  get the base addmittance, with a LN voltage base
+            Sbase = 1 # not yet pmd_data["baseMVA"] because this is done in _PMs.make_per_unit
+            Ybase_ln = Sbase/(pmd_data["basekv"]/sqrt(3))^2
+            # now convent b_cap to per unit
+            b_cap_pu = b_cap/Ybase_ln
+            #TODO clean this up when revisiting p.u. conversion
+            # this is needed because the shunt is non-dimensionalized at PM level later on
+            b_cap_pu = b_cap_pu
```
pseudocubic: Sorry, I'm not clear on the need for this line from the comment?

test/tp_opf.jl
```diff
@@ -246,31 +246,31 @@ end
         result = run_tp_opf(mp_data, PMs.ACPPowerModel, ipopt_solver)
 
         @test result["termination_status"] == PMs.LOCALLY_SOLVED
-        @test isapprox(result["objective"], 0.0597017; atol = 1e-4)
+        @test isapprox(result["objective"], 0.0599389; atol = 1e-4)
```
pseudocubic: Just to confirm, are these new values from OpenDSS?

src/io/opendss.jl
sanderclaeys: This is a typo, I moved what that line was supposed to do up. Fixing it now

PR #160, opened 2019-08-23
UPD: PowerModels v0.13 Compatible

pseudocubic:
Updates all types to use PMs v0.13 style, GenericPowerModel{T} eliminated. PMD types are updated to use PMs style, but names kept in their current form otherwise (Form -> Model, Models for types that are Unions). 

**We should discuss if any model names need to change before this is merged.**

Adds empty `"switch"` dictionary to data structure, will update in future PR to add switch parsing to OpenDSS parser.

Updates naming scheme of functions from _tp_ to _mc_ (by default, change later to make each more specific).

### TODO

- [ ] discuss and agree on PMD internal Model names (@ccoffrin, @sanderclaeys, @frederikgeth)
- [x] update documentation


src/core/constraint.jl
```diff
@@ -1,21 +1,21 @@
...
     for cnd in _PMs.conductor_ids(pm)
-        constraint_tp_theta_ref(pm, nw, cnd, i)
+        constraint_mc_theta_ref(pm, nw, cnd, i)
     end
 end
 
 
 "storage loss constraints"
-function constraint_tp_storage_loss(pm::_PMs.GenericPowerModel, n::Int, i, bus, r, x, standby_loss)
+function constraint_mc_storage_loss(pm::_PMs.AbstractPowerModel, n::Int, i, bus, r, x, standby_loss)
```
ccoffrin: Based on changes in the constraint template to use the function defined in PowerModels, this function should not be called and can be removed.

src/core/constraint_template.jl
```diff
@@ -182,58 +182,58 @@ function constraint_tp_flow_losses(pm::_PMs.GenericPowerModel, i::Int; nw::Int=p
...
     b_sh_to = branch["b_to"].values
 
-    constraint_tp_flow_losses(pm::_PMs.GenericPowerModel, nw, i, f_bus, t_bus, f_idx, t_idx, r, x, g_sh_fr, g_sh_to, b_sh_fr, b_sh_to)
+    constraint_mc_flow_losses(pm::_PMs.AbstractPowerModel, nw, i, f_bus, t_bus, f_idx, t_idx, r, x, g_sh_fr, g_sh_to, b_sh_fr, b_sh_to)
 end
 
 
 ""
-function constraint_tp_storage_exchange(pm::_PMs.GenericPowerModel, i::Int; nw::Int=pm.cnw)
+function constraint_mc_storage_exchange(pm::_PMs.AbstractPowerModel, i::Int; nw::Int=pm.cnw)
```
ccoffrin: Following PowerModels conventions, `storage_exchange` has been replaced by separate constraint templates for `constraint_storage_complementarity_nl` and `constraint_storage_loss`.  This can be posted as an issue to fix later if it is time consuming to fix here.

We should post an issue,

src/core/constraint_template.jl
```diff
@@ -182,58 +182,58 @@ function constraint_tp_flow_losses(pm::_PMs.GenericPowerModel, i::Int; nw::Int=p
...
     fixed = trans["fixed"]
     tm = trans["tm"]
-    constraint_tp_oltc_tap_fix(pm, i, fixed, tm)
+    constraint_mc_oltc_tap_fix(pm, i, fixed, tm)
 end
 
 
 "KCL including transformer arcs."
-function constraint_tp_power_balance_shunt_trans(pm::_PMs.GenericPowerModel, i::Int; nw::Int=pm.cnw, cnd::Int=pm.ccnd)
+function constraint_mc_power_balance_shunt_trans(pm::_PMs.AbstractPowerModel, i::Int; nw::Int=pm.cnw, cnd::Int=pm.ccnd)
```
ccoffrin: PowerModels v0.13 proposed a method for a "generic" power balance constraint.  We should post an issue to see if a similar design can be used in this package.

src/prob/tp_mld.jl
```diff
@@ -98,52 +98,52 @@ function post_tp_mld(pm::_PMs.GenericPowerModel)
...
 
     for i in _PMs.ids(pm, :bus), c in _PMs.conductor_ids(pm)
-        constraint_tp_power_balance_shunt_trans_shed(pm, i, cnd=c)
+        constraint_mc_power_balance_shunt_trans_shed(pm, i, cnd=c)
     end
 
     for i in _PMs.ids(pm, :storage)
         _PMs.constraint_storage_state(pm, i)
-        constraint_tp_storage_exchange(pm, i)
+        constraint_mc_storage_exchange(pm, i)
```
ccoffrin: I think the loss and complementarity constraints from PowerModels can be used directly here instead of calling this function.

src/core/constraint.jl
pseudocubic: Good catch! its definitely not called, or there's be an error, but I missed it when I was pruning functions.

src/core/constraint_template.jl
pseudocubic: Definitely I like the idea of making this function more generic, instead of having so many different variations for storage, loads, transformers, etc.

pseudocubic:
Because I have a lot of updates that depend on this I will merge this PR into the v0.6-rc branch. There will still be some time for comments on the names of the formulation types before the merging of v0.6-rc into master.

PR #172, opened 2019-09-11
FIX: OpenDSS parsing bugs

sanderclaeys:
There were some issues with the parser changes
1. When a bank has the same name as a transformer, the parser got stuck in a loop. I updated the banking code to prevent this, and fixed the merging as well (transformer_comp values are indexed on winding and conductor, not only conductor).
2. The branch phase reordering was not working as it should.

I am still investigating some other issues

### DIFF
--- a/src/io/dss_parse.jl
+++ b/src/io/dss_parse.jl
@@ -353,11 +353,11 @@ end
 
 
 "Reorders a `matrix` based on the order that phases are listed in on the from- (`pof`) and to-sides (`pot`)"
-function _reorder_matrix(matrix, pof, pot)
-    mat = deepcopy(matrix)
-    for (i, pf) in enumerate(pof)
-        for (j, pt) in enumerate(pot)
-            mat[i, j] = matrix[pf, pt]
+function _reorder_matrix(matrix, nconductors, phase_order)
+    mat = fill(0.0, nconductors, nconductors)
+    for (i, n) in enumerate(phase_order)
+        for (j, m) in enumerate(phase_order)
+            mat[n, m] = matrix[i, j]
         end
     end
     return mat
--- a/src/io/dss_parse.jl
+++ b/src/io/dss_parse.jl
@@ -355,9 +355,16 @@ end
 "Reorders a `matrix` based on the order that phases are listed in on the from- (`pof`) and to-sides (`pot`)"
 function _reorder_matrix(matrix, nconductors, phase_order)
     mat = fill(0.0, nconductors, nconductors)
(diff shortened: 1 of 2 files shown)


sanderclaeys:
@pseudocubic 

Another issue occurs when parsing for example
```
new transformer.reg1a phases=1 windings=2 bank=reg1 buses=(814.1 814r.1) conns='wye wye' kvs="14.376 14.376" kvas="2000 2000" XHL=.01
~ wdg=1 %r=.0001 wdg=2 %r=.0001 ppm=0
```

The transformer dictionary in `dss_data` for this component contains `"wdg"=>"2", "wdg_2"=>"2"`, and as a result the `"%rs"` property in `_create_transformer` is not populated correctly. I tried to find the cause, but this goes beyond my understanding of the dss parsing code.

pseudocubic:
@sanderclaeys can you please provide examples for the two claims you made that are broken so I can test this? 

As far as I can tell, none of the current tests should break when fixing those issues, we need to add additional tests that would be broken with the current parser that would be fixed with this PR

sanderclaeys:
@pseudocubic I will make a small example for the two that I fixed already. Can you have a look at what can cause this `"wdg"=>"2"` issue? That is quite self-contained.

I will go through the unit tests for the transformers that break now. I never checked them against the com results of OpenDSS, which have higher precision than the GUI results I checked against previously.

pseudocubic:
After some more investigating, I don't think reordering the phases truly has any effect other than reordering in the visualizations inside OpenDSS. I have tried the IEEE13 case with different settings a don't see an effect. I think the confusion comes from the reordering on the visualization, because at first glance I assumed there was an effect, but I was mistaken. Also, I cannot find any documentation to support the fact that reordering does anything either.

sanderclaeys:
My implementation of the reordering broke down when no conductors are supplied; I  fixed this now.

I will investigate tomorrow whether the order has an effect or not as well; the com interface gives more detail than the gui.

pseudocubic:
It looks like I'm not correct about phase reordering, its very subtle and apparently not visible on the user interface, it also seems to require a specially crafted case to see the differences.

I figured out the problem that you were having with the phase ordering, and have corrected it in v0.6-rc branch, along with the two other bugs you talked about (banking transformers and the %rs value not being correct when specifying %r separately by winding).

I will close this PR because while trying to figure this out I fixed all of these issues; if you discover there is still an issue with the phase ordering logic you should open a new PR based on the newest version of the v0.6-rc branch.

PR #186, opened 2019-11-02
sourcebus impedance rate_a

sanderclaeys:
Added a rate_a to the virtual branch modelling the internal impedance of the sourcebus. Also vbranches will now be created automatically with rate_a=Inf


src/io/opendss.jl
```diff
@@ -1082,6 +1082,16 @@ function _adjust_sourcegen_bounds!(pmd_data)
...
+
+    # set current rating of vbranch modelling internal impedance
+    sourcegen = pmd_data["gen"]["1"]
+    sourcebus = pmd_data["bus"][string(sourcegen["gen_bus"])]
+    sid = sourcebus["index"]
+    # the branch modelling the internal impedance, is the only branch connected to the sourcebus
+    vbranch = [br for (id, br) in pmd_data["branch"] if br["f_bus"]==sid || br["t_bus"]==sid][1]
+    vbranch["rate_a"] = _PMs.MultiConductorVector(fill(bound, length(vbranch["rate_a"])))
+    #vbranch["rate_b"] = bound
+    #vbranch["rate_c"] = bound
```
pseudocubic: you should search for the virtual branch by the name `"sourcebus_vbranch"`, so you won't need L1086-L1090, and you should remove L1093-L1094 since they don't do anything

src/io/opendss.jl
```diff
@@ -1231,6 +1244,7 @@ function _create_vbranch!(pmd_data, f_bus::Int, t_bus::Int; name="", source_id="
     end
     vbranch["angmin"] = -_PMs.MultiConductorVector(ones(ncnd))*60
     vbranch["angmax"] = _PMs.MultiConductorVector(ones(ncnd))*60
+    vbranch["rate_a"] = rate_a
```
pseudocubic: Instead of creating a new `rate_a` kwarg with a default that's complex like that, I would use `get(kwargs, :rate_a, rate_a_default)` on L1247.

sanderclaeys: I like that pattern, I should use it more.

sanderclaeys:
@pseudocubic I adressed your comments and added a unit test

frederikgeth:
Can this be merged? The test failures seem not to be the fault of the implementation

PR #190, opened 2019-12-12
scs fix

sanderclaeys:
Changing the SCS solver by
- adding the setting `alpha=1.8`
- decreasing the tolerance `eps=1e-5`,
- and increasing the maximum iterations `max_iters=20000`,

makes all tests pass again as before. If total run time becomes an issue, we can increase the tolerance to `eps=1e-4` and increase the `atol` on the affected tests instead.

@ccoffrin adding `alpha=1.8` did the trick, without that it would max out on 1000k iterations for `eps=1e-4` on some tests.


ccoffrin:
@pseudocubic, LGTM just make sure to squash merge.

PR #195, opened 2020-01-22
Fix for ACR OPF

frederikgeth:
@ccoffrin, is this going in the right direction? More tests are passing.

- parses model and calls ipopt
- returns similar result to ACP
- delta loads / transformers not yet tested

### DIFF
--- a/src/form/acr.jl
+++ b/src/form/acr.jl
@@ -97,7 +97,6 @@ function constraint_mc_power_balance_slack(pm::_PMs.AbstractACRModel, nw::Int, i
     cstr_q = []
 
     for c in _PMs.conductor_ids(pm; nw=nw)
-        JuMP.@constraint(pm.model,
             sum(p[a][c] for a in bus_arcs)
             + sum(psw[a_sw][c] for a_sw in bus_arcs_sw)
             + sum(pt[a_trans][c] for a_trans in bus_arcs_trans)
@@ -108,7 +107,9 @@ function constraint_mc_power_balance_slack(pm::_PMs.AbstractACRModel, nw::Int, i
             - sum(gs[c] for gs in values(bus_gs))*(vr[c]^2 + vi[c]^2)
             + p_slack[c]
         )
-        JuMP.@constraint(pm.model,
+        push!(cstr_p, cp)
+
+        cq = JuMP.@constraint(pm.model,
             sum(q[a][c] for a in bus_arcs)
             + sum(qsw[a_sw][c] for a_sw in bus_arcs_sw)
             + sum(qt[a_trans][c] for a_trans in bus_arcs_trans)
@@ -119,11 +120,12 @@ function constraint_mc_power_balance_slack(pm::_PMs.AbstractACRModel, nw::Int, i
             + sum(bs[c] for bs in values(bus_bs))*(vr[c]^2 + vi[c]^2)
             + q_slack[c]
         )
+        push!(cstr_q, cq)
     end
 
     if _PMs.report_duals(pm)
-        _PMs.sol(pm, nw, :bus, i)[:lam_kcl_r] = cstr_p
-        _PMs.sol(pm, nw, :bus, i)[:lam_kcl_i] = cstr_q
+        _PMs.sol(pm, nw, :bus, i)[:lam_kcl_p] = cstr_p
+        _PMs.sol(pm, nw, :bus, i)[:lam_kcl_q] = cstr_q
     end
 end
 
@@ -147,7 +149,7 @@ function constraint_mc_power_balance(pm::_PMs.AbstractACRModel, nw::Int, i::Int,
     cstr_q = []
 
     for c in _PMs.conductor_ids(pm; nw=nw)
-        JuMP.@constraint(pm.model,
+        cp = JuMP.@constraint(pm.model,
             sum(p[a][c] for a in bus_arcs)
             + sum(psw[a_sw][c] for a_sw in bus_arcs_sw)
             + sum(pt[a_trans][c] for a_trans in bus_arcs_trans)
@@ -157,7 +159,9 @@ function constraint_mc_power_balance(pm::_PMs.AbstractACRModel, nw::Int, i::Int,
             - sum(pd[c] for pd in values(bus_pd))
             - sum(gs[c] for gs in values(bus_gs))*(vr[c]^2 + vi[c]^2)
         )
-        JuMP.@constraint(pm.model,
+        push!(cstr_p, cp)
+
+        cq = JuMP.@constraint(pm.model,
             sum(q[a][c] for a in bus_arcs)
             + sum(qsw[a_sw][c] for a_sw in bus_arcs_sw)
             + sum(qt[a_trans][c] for a_trans in bus_arcs_trans)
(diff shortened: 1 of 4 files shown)


src/form/acr.jl
```diff
@@ -18,162 +18,211 @@ function variable_mc_voltage(pm::_PMs.AbstractACRModel; nw=pm.cnw, kwargs...)
...
+    qsw  = get(_PMs.var(pm, nw),  :qsw, Dict()); _PMs._check_var_keys(qsw, bus_arcs_sw, "reactive power", "switch")
+    pt   = get(_PMs.var(pm, nw),   :pt, Dict()); _PMs._check_var_keys(pt, bus_arcs_trans, "active power", "transformer")
+    qt   = get(_PMs.var(pm, nw),   :qt, Dict()); _PMs._check_var_keys(qt, bus_arcs_trans, "reactive power", "transformer")
+    p_slack = _PMs.var(pm, nw, :p_slack, i)
+    q_slack = _PMs.var(pm, nw, :q_slack, i)
+
+    cstr_p = []
+    cstr_q = []
+
+    for c in _PMs.conductor_ids(pm; nw=nw)
```
ccoffrin: cstr_p and cstr_q are not populated in this loop.

src/form/acr.jl
```diff
@@ -18,162 +18,211 @@ function variable_mc_voltage(pm::_PMs.AbstractACRModel; nw=pm.cnw, kwargs...)
...
+            - sum(qs[s][c] for s in bus_storage)
+            - sum(qd[c] for qd in values(bus_qd))
+            + sum(bs[c] for bs in values(bus_bs))*(vr[c]^2 + vi[c]^2)
+            + q_slack[c]
+        )
+    end
+
+    if _PMs.report_duals(pm)
+        _PMs.sol(pm, nw, :bus, i)[:lam_kcl_r] = cstr_p
+        _PMs.sol(pm, nw, :bus, i)[:lam_kcl_i] = cstr_q
```
ccoffrin: The standard name here so far is `_p` and `_q`.

src/form/acr.jl
```diff
@@ -18,162 +18,211 @@ function variable_mc_voltage(pm::_PMs.AbstractACRModel; nw=pm.cnw, kwargs...)
...
-        sum(q[a] for a in bus_arcs)
-        + sum(qsw[a_sw] for a_sw in bus_arcs_sw)
-        + sum(qt[a_trans] for a_trans in bus_arcs_trans)
-        ==
-        sum(qg[g] for g in bus_gens)
-        - sum(qs[s] for s in bus_storage)
-        - sum(qd for qd in values(bus_qd))
-        + sum(bs for bs in values(bus_bs))*(vr^2 + vi^2)
-    )
+function constraint_mc_power_balance(pm::_PMs.AbstractACRModel, nw::Int, i::Int, bus_arcs, bus_arcs_sw, bus_arcs_trans, bus_gens, bus_storage, bus_pd, bus_qd, bus_gs, bus_bs)
```
ccoffrin: Same notes as `_slack` variant.

src/form/acr.jl
```diff
@@ -18,162 +18,211 @@ function variable_mc_voltage(pm::_PMs.AbstractACRModel; nw=pm.cnw, kwargs...)
...
-        sum(q[a] for a in bus_arcs)
-        + sum(qsw[a_sw] for a_sw in bus_arcs_sw)
-        + sum(qt[a_trans] for a_trans in bus_arcs_trans)
-        ==
-        sum(qg[g] for g in bus_gens)
-        - sum(qs[s] for s in bus_storage)
-        - sum(qd[l] for l in bus_loads)
-        + sum(bs for bs in values(bus_bs))*(vr^2 + vi^2)
-    )
+function constraint_mc_power_balance_load(pm::_PMs.AbstractACRModel, nw::Int, i::Int, bus_arcs, bus_arcs_sw, bus_arcs_trans, bus_gens, bus_storage, bus_loads, bus_gs, bus_bs)
```
ccoffrin: Same notes as slack variant.

src/form/acr.jl
```diff
@@ -199,34 +248,37 @@ q_fr ==  sum(
...
-                                            -vr_fr[c]*(b[c,d]*(vr_fr[d]-vr_to[d])+g[c,d]*(vi_fr[d]-vi_to[d]))
-                                            +vi_fr[c]*(g[c,d]*(vr_fr[d]-vr_to[d])-b[c,d]*(vi_fr[d]-vi_to[d]))
-                                            for d in _PMs.conductor_ids(pm))
-                                          # the shunt element is identical but vr_to=vi_to=0
-                                          + sum(
-                                            -vr_fr[c]*(b_fr[c,d]*vr_fr[d]+g_fr[c,d]*vi_fr[d])
-                                            +vi_fr[c]*(g_fr[c,d]*vr_fr[d]-b_fr[c,d]*vi_fr[d])
-                                            for d in _PMs.conductor_ids(pm))
-    )
+function constraint_mc_ohms_yt_from(pm::_PMs.AbstractACRModel, n::Int, f_bus, t_bus, f_idx, t_idx, g, b, g_fr, b_fr, tr, ti, tm)
```
ccoffrin: Indentation could probably be improved for readability.

src/form/acr.jl
```diff
@@ -238,33 +290,39 @@ p[t_idx] ==  (g+g_to)*v[t_bus]^2 + (-g*tr-b*ti)/tm*(v[t_bus]*v[f_bus]*cos(t[t_bu
...
 
 ""
-function constraint_load_power_wye(pm::_PMs.AbstractACRModel, nw::Int, cnd::Int, load_id::Int, pd::Real, qd::Real)
-    _PMs.var(pm, nw, cnd, :pd)[load_id] = pd
-    _PMs.var(pm, nw, cnd, :qd)[load_id] = qd
+function constraint_load_power_wye(pm::_PMs.AbstractACRModel, nw::Int, load_id::Int, pd::Real, qd::Real)
+    for c in _PMs.conductor_ids(pm; nw=nw)
+        _PMs.var(pm, nw, :pd)[load_id][c] = pd[c]
+        _PMs.var(pm, nw, :qd)[load_id][c] = qd[c]
+    end
```
ccoffrin: I don't yet have a good sense if intermediate expressions should be in the solution dict or not.  At least thus far, I the solution dict only contains model variables and constraint duals, not intermediate expressions.

Still, becouse you are populating `var` here, you could also consider adding the `:pd` terms to the solution as well.

Thoughts?

src/form/acr.jl
frederikgeth: ok, don't forget to check this in PowerModels.jl then, I think they're all like that at the moment.

src/form/acr.jl
frederikgeth: indeed, neither were they in ACP apparently. Is it ok to use push!?

src/form/acr.jl
```diff
@@ -18,162 +18,220 @@ function variable_mc_voltage(pm::_PMs.AbstractACRModel; nw=pm.cnw, kwargs...)
...
+            + sum(psw[a_sw][c] for a_sw in bus_arcs_sw)
+            + sum(pt[a_trans][c] for a_trans in bus_arcs_trans)
+            ==
+            sum(pg[g][c] for g in bus_gens)
+            - sum(ps[s][c] for s in bus_storage)
+            - sum(pd[c] for pd in values(bus_pd))
+            - sum(gs[c] for gs in values(bus_gs))*(vr[c]^2 + vi[c]^2)
+            + p_slack[c]
+        )
+        push!(cstr_p, cp)
```
frederikgeth: push! like this?

src/form/acr.jl
frederikgeth: done

src/form/acr.jl
frederikgeth: I'm not sure what to think of it either at this moment

src/form/acr.jl
ccoffrin: ok my bad there! `push!` should be fine.  The type the array won't be strongly typed but that should be fine for now.

src/form/acr.jl
ccoffrin: You are correct.  PowerModels was using a mix of `lam_kcl_r` and `kcl_p`.  Lets stick to what you are using `lam_kcl_r`. I think it will generalize better.

src/form/acr.jl
ccoffrin: Agreed.  For now let's adopt the convention only true variables that occur in the JuMP model variables are put in the solution by default.

frederikgeth:
279 tests passing now (OPF + PF / ACP + ACR), will merge and then continue debugging

PR #204, opened 2020-01-25
fix all storage tests

frederikgeth:


frederikgeth:
772 passed, 89 failed, 98 errored, 0 broken.

PR #205, opened 2020-01-25
Pms v15 trans

sanderclaeys:
transformers and load models fixed (+115)


sanderclaeys:
also fixed the final test of pf.jl, related to SOCWRPowermModel

PR #211, opened 2020-01-28
Pms v15 fixes

frederikgeth:
numerous fixes

### DIFF
--- a/src/form/apo.jl
+++ b/src/form/apo.jl
@@ -6,17 +6,17 @@ end
 
 
 "apo models ignore reactive power flows"
-function variable_mc_generation_on_off_reactive(pm::_PMs.AbstractActivePowerModel; kwargs...)
+function variable_mc_reactive_generation_on_off(pm::_PMs.AbstractActivePowerModel; kwargs...)
 end
 
 
 "apo models ignore reactive power flows"
-function variable_mc_reactive_storage(pm::_PMs.AbstractActivePowerModel; kwargs...)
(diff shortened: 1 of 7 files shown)


frederikgeth:
Some tests did not pass: 1085 passed, 83 failed, 14 errored, 0 broken.

src/form/apo.jl
```diff
@@ -6,17 +6,17 @@ end
 
 
 "apo models ignore reactive power flows"
-function variable_mc_generation_on_off_reactive(pm::_PMs.AbstractActivePowerModel; kwargs...)
+function variable_mc_reactive_generation_on_off(pm::_PMs.AbstractActivePowerModel; kwargs...)
```
sanderclaeys: reactive at the end?

src/form/apo.jl
```diff
@@ -277,14 +288,18 @@ function constraint_storage_loss(pm::_PMs.AbstractActivePowerModel, n::Int, i, b
 end
 
 function constraint_mc_storage_on_off(pm::_PMs.AbstractActivePowerModel, n::Int, i, pmin, pmax, qmin, qmax, charge_ub, discharge_ub)
+    @assert all(pmin.<=pmax)
```
sanderclaeys: Better to add this test at the template level and do it at once for all formulations?

src/form/apo.jl
frederikgeth: the existing ones had this order?

src/form/apo.jl
frederikgeth: remove

PR #213, opened 2020-01-29
Pms v15 bugfixes

frederikgeth:
no errors remaining :)


frederikgeth:
1052 passed, 57 failed, 0 errored, 0 broken.

PR #215, opened 2020-01-29
v8-rc

sanderclaeys:
```
Test Summary:           | Pass  Total
PowerModelsDistribution |  634    634
```


sanderclaeys:
I updated the version of InfrastructureModels to 0.4, else I kept on getting conflicting requirements. However, I think we now should also update _value2string to value2string?

src/core/constraint_template.jl
```diff
@@ -101,7 +121,7 @@ end
 
 
 ""
-function constraint_mc_model_current(pm::AbstractUBFModels; nw::Int=pm.cnw)
+function constraint_mc_model_current(pm::Union{AbstractUBFModels,LPLinUBFPowerModel}; nw::Int=pm.cnw)
```
pseudocubic: This seems like it should be solved in src/core/types.jl since it seems natural that LPLinUBFPowerModel should be under AbstractUBFModels. If that is not true, the formulation needs to be renamed to avoid confusion

src/core/solution.jl
```diff
@@ -1,431 +1,445 @@
 import LinearAlgebra: tr
```
pseudocubic: There are tons of lines commented out in this file. Do you foresee any of it being necessary down the line? If not, go ahead and delete it.

src/core/variable.jl
```diff
@@ -1,210 +1,472 @@
...
+            JuMP.set_lower_bound(w[i][c], bus["vmin"][c]^2)
+            JuMP.set_upper_bound(w[i][c], bus["vmax"][c]^2)
         end
+    end
+
+    report && _PMs.sol_component_value(pm, nw, :bus, :w, _PMs.ids(pm, nw, :bus), w)
+end
+
+#
+# ""
```
pseudocubic: I assume this must be for SOCWR? Do you think you will reuse any of this function when you rewrite SOCWR? If not go ahead and remove

test/pf.jl
```diff
@@ -1,5 +1,8 @@
 @info "running power flow (pf) tests"
 
+calc_vm_w(result, id) = sqrt.(      result["solution"]["bus"][id]["w"])
+calc_vm_W(result, id) = sqrt.(diag( result["solution"]["bus"][id]["Wr"]))
```
pseudocubic: Should be in test/common.jl

test/pf.jl
```diff
@@ -37,19 +40,19 @@
...
-        end
-
-        @testset "5-bus coupled meshed network (b) soc pf" begin
-            result = run_mc_pf("../test/data/matlab/case5_c_m_b.m", PMs.SOCWRPowerModel, ipopt_solver)
+        # @testset "5-bus coupled meshed network (b) acr pf" begin
+        #     result = PMD.run_mc_pf("../test/data/matlab/case5_c_m_b.m", PMs.ACRPowerModel, ipopt_solver)
+        #
+        #     @test result["termination_status"] == PMs.LOCALLY_SOLVED
+        #     @test isapprox(result["objective"], 0.0; atol=1e-2)
+        # end
```
pseudocubic: Reason this is disabled? It might be helpful to put very brief comments above commented-out code to make it easier to track

test/opf_bf.jl
```diff
@@ -1,5 +1,10 @@
 @info "running branch-flow optimal power flow (opf_bf) tests"
 
+calc_vm_w(result, id) = sqrt.(      result["solution"]["bus"][id]["w"])
+calc_vm_W(result, id) = sqrt.(diag( result["solution"]["bus"][id]["Wr"]))
+
```
pseudocubic: should be in test/common.jl

test/opf.jl
```diff
@@ -2,64 +2,66 @@
...
-            PMD.make_multiconductor!(mp_data, 3)
-            result = run_mc_opf(mp_data, PMs.ACRPowerModel, ipopt_solver)
-
-            @test result["termination_status"] == PMs.LOCALLY_SOLVED
-            @test isapprox(result["objective"], 614.007; atol=1e-1)
-
-            @test all(isapprox(result["solution"]["gen"]["1"]["pg"][c],  2.192189; atol=1e-3) for c in 1:mp_data["conductors"])
-            @test all(isapprox(result["solution"]["bus"]["2"]["va"][c], PMD._wrap_to_pi(-0.071853+2*pi/mp_data["conductors"]*(1-c)); atol=1e-4) for c in 1:mp_data["conductors"])
-        end
+        # @testset "5-bus matpower acp opf" begin
```
pseudocubic: Comments with reason for disabling all of these tests would really help

test/data/pti/parser_test_defaults.raw
```diff
@@ -0,0 +1,50 @@
+,,,,,
```
pseudocubic: We can use files in dependent packages without copying them over, we shouldn't add these files directly like this

src/prob/test.jl
```diff
@@ -4,70 +4,363 @@
...
+#             constraint_thermal_limit_to(pm, i, cnd=c)
+#         end
+#
+#         for i in ids(pm, :dcline)
+#             constraint_dcline(pm, i, cnd=c)
+#         end
+#     end
+#
+#     _PMs.objective_min_fuel_cost(pm)
+# end
```
pseudocubic: This seems like something we can just drop, instead of leaving it commented out

src/form/wr.jl
```diff
@@ -1,56 +1,58 @@
...
-        + sum(qt[a_trans] for a_trans in bus_arcs_trans)
-        ==
-        sum(qg[g] for g in bus_gens)
-        - sum(qs[s] for s in bus_storage)
-        - sum(qd for qd in values(bus_qd))
-        + sum(bs for bs in values(bus_bs))*w
-    )
-end
+# "power balance constraint with line shunts and transformers for relaxed WR forms"
+# function constraint_mc_power_balance(pm::_PMs.AbstractWRModel, nw::Int, c::Int, i::Int, bus_arcs, bus_arcs_sw, bus_arcs_trans, bus_gens, bus_storage, bus_pd, bus_qd, bus_gs, bus_bs)
```
pseudocubic: Intend to reuse, or will completely rewrite?

src/core/variable.jl
sanderclaeys: I removed all SOCWR related stuff that will become obsolete when we add it again (tests, variable creation). I kept the constraints however in comments, as those will probably be largely recycled.

test/data/pti/parser_test_defaults.raw
sanderclaeys: @frederikgeth I don't see where that file is being used, can I delete it?

src/form/wr.jl
sanderclaeys: See previous comment on SOCWR, constraints only will be reused.

sanderclaeys:
@pseudocubic I will make the small changes you requested above. In the meanwhile, can you approve #216 ? This is the last piece of changes of the refactoring for PMD v0.8; once this is merged into pms-v15, I can create a new feature branch pmd-v9 where we will merge all the new functionality we worked on (matrix bus shunts, IVR, data model, 4-wire etc.)

src/core/constraint_template.jl
sanderclaeys: I ended up cleaning up the linearizations in #217 , once Frederik reviewed this I will pull it in this branch.

src/PowerModelsDistribution.jl
```diff
@@ -13,16 +14,14 @@ module PowerModelsDistribution
     function __init__()
         global _LOGGER = Memento.getlogger(PowerModels)
     end
-
+
```
pseudocubic: Looks like there's some leftover whitespace?

test/matlab.jl
```diff
@@ -10,11 +10,11 @@
...
         @test length(data["branch"]) == 4
 
-        @test isa(data["bus"]["1"]["vm"], PMs.MultiConductorVector{Float64})
-        @test isa(data["load"]["1"]["pd"], PMs.MultiConductorVector{Float64})
-        @test isa(data["gen"]["1"]["pg"], PMs.MultiConductorVector{Float64})
-        @test isa(data["branch"]["1"]["b_fr"], PMs.MultiConductorMatrix{Float64})
-        @test isa(data["branch"]["1"]["br_x"], PMs.MultiConductorMatrix{Float64})
+        @test isa(data["bus"]["1"]["vm"], PMD.Vector{Float64})
+        @test isa(data["load"]["1"]["pd"], PMD.Vector{Float64})
+        @test isa(data["gen"]["1"]["pg"], PMD.Vector{Float64})
```
pseudocubic: Still some lines with PMD.Vector instead of Vector remaining

test/pf.jl
sanderclaeys: I dropped all matlab format related code and tests, and restored the matpower related tests. We want to keep the functionality to import balanced matpower cases, but the custom matlab format will be dropped in favor of the one that matpower will develop. Therefore, better to drop it now and not spend more time on the tests related to that.

frederikgeth:
I think the problem type opf_bctr is untested (and broken). What is the purpose of this problem definition?

pseudocubic:
@frederikgeth After looking closely at the git history, all I can guess is that it was somehow an initial version of a problem with transformers that was left over and therefore never updated as we continued to upgrade PMD. Go ahead and remove it.

PR #216, opened 2020-01-31
rm MultiConductor

sanderclaeys:
removed MultiConductorVector/MultiConductorMatrix, and updated the parser and the rest of the code to reflect this


src/core/multiconductor.jl
```diff
@@ -1,206 +1,5 @@
...
-
-
-conductor_value(mc::Any, conductor::Int) = mc
-conductor_value(mc::Any, conductor_i::Int, conductor_j::Int) = mc
-conductor_value(mc::MultiConductorVector, conductor::Int) = mc[conductor]
-conductor_value(mc::MultiConductorMatrix{T}, conductor::Int) where T = MultiConductorVector{T}(mc[conductor])
-conductor_value(mc::MultiConductorMatrix, conductor_i::Int, conductor_j::Int) = mc[conductor_i, conductor_j]
-
 
 "https://stackoverflow.com/questions/39039553/lower-triangular-matrix-in-julia"
```
pseudocubic: Seems like these remaining functions could move to src/core/data.jl

test/matlab.jl
```diff
@@ -28,6 +28,6 @@
         @test length(data["gen"]) == 5
         @test length(data["branch"]) == 4
 
-        @test isa(data["shunt"]["1"]["bs"], PMD.MultiConductorVector{Float64})
+        @test isa(data["shunt"]["1"]["bs"], PMD.Vector{Float64})
```
pseudocubic: Seems like it should just be `Vector` instead of `PMD.Vector`

src/io/json.jl
```diff
@@ -52,13 +53,45 @@ function parse_json(io::Union{IO,String}; kwargs...)::Dict{String,Any}
...
+end
+
+
+function show_json(io::StructuralContext, ::PMDSerialization, f::Matrix{<:Any})
+    N, M = size(f)
+    value = string("[", join([join([f[i,j] for j in 1:M], " ") for i in 1:N], "; "), "]")
+    eltyp = isempty(f) ? eltype(f) : typeof(f[1,1])
+        out = Dict(:type=>:Matrix, :eltype=>eltyp, :value=>value)
+    return show_json(io, StandardSerialization(), out)
+end
```
pseudocubic: Just want to confirm I'm reading this right, matrices turn into `[a b c; d e f; g h i]` in the json string?

test/multiconductor.jl
```diff
@@ -62,17 +62,17 @@ calc_va_acr(result, id) = angle.(result["solution"]["bus"][id]["vr"] +im* result
...
 
         mc_data = build_mc_data!("../test/data/matpower/case5.m")
-        mc_data["gen"]["1"]["pmax"] = PMD.MultiConductorVector([Inf, Inf, Inf])
-        mc_data["gen"]["1"]["qmin"] = PMD.MultiConductorVector([-Inf, -Inf, -Inf])
-        mc_data["gen"]["1"]["bool_test"] = PMD.MultiConductorVector([true, true, false])
-        mc_data["gen"]["1"]["string_test"] = PMD.MultiConductorVector(["a", "b", "c"])
+        mc_data["gen"]["1"]["pmax"] = [Inf, Inf, Inf]
+        mc_data["gen"]["1"]["qmin"] = [-Inf, -Inf, -Inf]
+        mc_data["gen"]["1"]["bool_test"] = [true, true, false]
+        mc_data["gen"]["1"]["string_test"] = ["a", "b", "c"]
```
pseudocubic: I think we can delete this file now

src/io/json.jl
sanderclaeys: Correct. I couldn't simply print the matrices because then they are not printed with full precision, leading to rounding of the numbers when you saved them.

PR #247, opened 2020-02-19
ADD: LoadShape parsing

pseudocubic:
Add capability to parse LoadShapes from DSS files, with support for
binary files with single or double precision, and csv files.

OpenDSS supports a syntax that enables loading from files series data
into any field that is an array. The parser has been updated to support
this feature.

Loadshapes are not yet supported natively into the PowerModelsDistribution
data structure, this feature add only enables native parsing of LoadShape
structures into a dictionary format. Full support for expanding LoadShape
data into multinetworks will be added in a future pull request.

Adds Unit tests and updates changelog


ccoffrin:
See the time series object in InfrastructureModels for the standard way to
integrate into the PM data format.

PR #250, opened 2020-02-21
optional bounds

sanderclaeys:
This is a PR in preparation of the data model PR, which
- updates the transformer data model in the mathematical model;
- makes all bounds optional. When `bounded=true`, it will still check whether the bounds are present in the data model; if they are not, then the bounds are not applied. This is usually checked by a method in `data.jl`, which returns a vector of `Inf` when the bound is not present, which is subsequently ignored by the new wrapper functions to set the bounds.

Also, I updated several OpenDSS case files. These use the syntax `bus1=x.1.2.3.0` for a line which has `phases=3`, i.e. only 3 conductors! Therefore, this should be `bus1=x.1.2.3`. Perhaps this works because OpenDSS silently ignores excess connections, but this is not correct.


ccoffrin:
Regarding this change,

> renames a number of bounds (rate_a->s_rating and c_rating_a->c_rating);

I think these have to remain the same so that we have consistency with the PowerModels data model.  And I would not recomend making local variable names deviate from the data model names.

sanderclaeys:
@ccoffrin, @frederikgeth We are going more and more to a current-limit first model, so I would like to make it explicit that a bound refers to power, i.e., adding 's_' to the power limits. Though I don't have a problem with making an exception for the branches because of legacy and PowerModels convention.

What about new components and bounds, i.e. current and power limits for transformers, which don't have a counterpart in PowerModels? Ideally I would drop there the a/b/c suffix, as the b/c variant is never used in our models and they are not exactly very descriptive.

ccoffrin:
I agree `s_` and `c_` are the way to go in the future.  But PM and PMD should be kept consistent until we can make those name changes at the same time (or nearly so).  For new fields in PMD there is more flexibility, however I would not recomend having conventions that are not internally consistent.  Doing a big rename all at the same time is not generally a big effort.

sanderclaeys:
@ccoffrin Okay, then I will revert to `rate_a` and `c_rating_a` for both branches and transformers, with the intention to update it all at once in the future.

test/data/opendss/test2_master.dss
```diff
@@ -15,7 +15,7 @@ New Line.L1-2 bus1=b3-1.2 bus2=b4.2 length=0.032175613 units=km Linecode=lc1
 New Line.L2 bus1=b5 bus2=b6_check-chars length=0.013516796 units=none linecode=lc/2
 New "Line.L3" phases=3 bus1=b7.1.2.3 bus2=b9.1.2.3 linecode=300 normamps=400 emergamps=600 faultrate=0.1 pctperm=20 repair=0 length=2.58
 New "Line._L4" phases=3 bus1=b8.1.2.3 bus2=b10.1.2.3 linecode=300 normamps=400 emergamps=600 faultrate=0.1 pctperm=20 repair=0 length=1.73 switch=y
-New line.l5 phases=3 bus1=_b2.1.2.3.0 bus2=b7.1.2.3.0 linecode=lc8
+New line.l5 phases=3 bus1=_b2.1.2.3 bus2=b7.1.2.3 linecode=lc8
```
pseudocubic: I would leave this one as is, it is a test for the parser itself, not a real case

test/runtests.jl
```diff
@@ -26,7 +26,7 @@ using LinearAlgebra
 pms_path = joinpath(dirname(pathof(PowerModels)), "..")
 pmd_path = joinpath(dirname(pathof(PowerModelsDistribution)), "..")
 
-ipopt_solver = with_optimizer(Ipopt.Optimizer, tol=1e-6, print_level=0)
+ipopt_solver = with_optimizer(Ipopt.Optimizer, tol=1e-7, print_level=0)
```
pseudocubic: Reason for the change in tolerance?

src/core/variable.jl
```diff
@@ -81,9 +87,11 @@ function variable_mc_voltage_imaginary(pm::_PMs.AbstractPowerModel; nw::Int=pm.c
     )
 
     if bounded
-        for (i,bus) in _PMs.ref(pm, nw, :bus), c in cnds
-            JuMP.set_lower_bound(vi[i][c], -bus["vmax"][c])
-            JuMP.set_upper_bound(vi[i][c],  bus["vmax"][c])
+        for (i,bus) in _PMs.ref(pm, nw, :bus)
+            if haskey(bus, "vmax")
+                JuMP.set_lower_bound.(vi[i], -bus["vmax"])
+                JuMP.set_upper_bound.(vi[i],  bus["vmax"])
```
pseudocubic: should we instead be using the internal `set_lower_bound` and `set_upper_bound` here?

src/core/variable.jl
```diff
@@ -60,9 +64,11 @@ function variable_mc_voltage_real(pm::_PMs.AbstractPowerModel; nw::Int=pm.cnw, b
     )
 
     if bounded
-        for (i,bus) in _PMs.ref(pm, nw, :bus), c in cnds
-            JuMP.set_lower_bound(vr[i][c], -bus["vmax"][c])
-            JuMP.set_upper_bound(vr[i][c],  bus["vmax"][c])
+        for (i,bus) in _PMs.ref(pm, nw, :bus)
+            if haskey(bus, "vmax")
+                JuMP.set_lower_bound.(vr[i], -bus["vmax"])
+                JuMP.set_upper_bound.(vr[i],  bus["vmax"])
```
pseudocubic: should we instead be using the internal `set_lower_bound` and `set_upper_bound` here?

src/core/variable.jl
```diff
@@ -150,18 +150,10 @@ function variable_mc_branch_flow_reactive(pm::_PMs.AbstractPowerModel; nw::Int=p
...
-                    JuMP.set_lower_bound(q[arc][c], flow_lb[l])
-                end
-                if !isinf(flow_ub[l])
-                    JuMP.set_upper_bound(q[arc][c], flow_ub[l])
-                end
-            end
+        for (l,i,j) in _PMs.ref(pm, nw, :arcs)
+            smax = _calc_branch_power_max(_PMs.ref(pm, nw, :branch, l), _PMs.ref(pm, nw, :bus, i))
+            JuMP.set_upper_bound.(q[(l,i,j)],  smax)
+            JuMP.set_lower_bound.(q[(l,i,j)], -smax)
```
pseudocubic: should we instead be using the internal `set_lower_bound` and `set_upper_bound` here?

src/core/variable.jl
```diff
@@ -612,14 +527,14 @@ function variable_mc_transformer_flow_reactive(pm::_PMs.AbstractPowerModel; nw::
...
-                JuMP.set_upper_bound(qt[arc][c], flow_ub[c])
-            end
+        for arc in _PMs.ref(pm, nw, :arcs_from_trans)
+            (t,i,j) = arc
+            rate_a_fr, rate_a_to = _calc_transformer_power_ub_frto(_PMs.ref(pm, nw, :transformer, t), _PMs.ref(pm, nw, :bus, i), _PMs.ref(pm, nw, :bus, j))
+
+            JuMP.set_lower_bound.(qt[(t,i,j)], -rate_a_fr)
+            JuMP.set_upper_bound.(qt[(t,i,j)],  rate_a_fr)
+            JuMP.set_lower_bound.(qt[(t,j,i)], -rate_a_fr)
+            JuMP.set_upper_bound.(qt[(t,j,i)],  rate_a_fr)
```
pseudocubic: should we instead be using the internal `set_lower_bound` and `set_upper_bound` here?

src/core/variable.jl
```diff
@@ -888,9 +803,13 @@ function variable_mc_generation_active(pm::_PMs.AbstractPowerModel; nw::Int=pm.c
...
     if bounded
-        for (i,gen) in _PMs.ref(pm, nw, :gen), c in cnds
-            JuMP.set_lower_bound(pg[i][c], gen["pmin"][c])
-            JuMP.set_upper_bound(pg[i][c], gen["pmax"][c])
+        for (i,gen) in _PMs.ref(pm, nw, :gen)
+            if haskey(gen, "pmin")
+                JuMP.set_lower_bound.(pg[i], gen["pmin"])
+            end
+            if haskey(gen, "pmax")
+                JuMP.set_upper_bound.(pg[i], gen["pmax"])
```
pseudocubic: should we instead be using the internal `set_lower_bound` and `set_upper_bound` here?

src/core/variable.jl
```diff
@@ -911,9 +830,13 @@ function variable_mc_generation_reactive(pm::_PMs.AbstractPowerModel; nw::Int=pm
...
     if bounded
-        for (i,gen) in _PMs.ref(pm, nw, :gen), c in cnds
-            JuMP.set_lower_bound(qg[i][c], gen["qmin"][c])
-            JuMP.set_upper_bound(qg[i][c], gen["qmax"][c])
+        for (i,gen) in _PMs.ref(pm, nw, :gen)
+            if haskey(gen, "qmin")
+                JuMP.set_lower_bound.(qg[i], gen["qmin"])
+            end
+            if haskey(gen, "qmax")
+                JuMP.set_upper_bound.(qg[i], gen["qmax"])
```
pseudocubic: should we instead be using the internal `set_lower_bound` and `set_upper_bound` here?

src/form/dcp.jl
```diff
@@ -81,17 +81,11 @@ function variable_mc_branch_flow_active(pm::_PMs.AbstractAPLossLessModels; nw::I
...
-                    JuMP.set_lower_bound(p[arc][cnd], flow_lb[l])
-                end
-                if !isinf(flow_ub[l])
-                    JuMP.set_upper_bound(p[arc][cnd], flow_ub[l])
-                end
+        for (l,i,j) in _PMs.ref(pm, nw, :arcs_from)
+            smax = _calc_branch_power_max(_PMs.ref(pm, nw, :branch, l), _PMs.ref(pm, nw, :bus, i))
+            if !ismissing(smax)
+                JuMP.set_upper_bound.(p[(l,i,j)],  smax)
+                JuMP.set_lower_bound.(p[(l,i,j)], -smax)
```
pseudocubic: should we instead be using the internal `set_lower_bound` and `set_upper_bound` here?

test/runtests.jl
sanderclaeys: This solved some issues I was having. Though after investigsting it more, it turned out some test from your last commit got overwritten I believe, causing the issue. 

In any case, this is not needed anymore, I removed it.

PR #254, opened 2020-03-12
Update to JuMP v0.21

ccoffrin:
@pseudocubic, what do you think about merging this now vs waiting for a fix in JuMP?

I am inclined to tag a new version that supports JuMP v0.21.


ccoffrin:
Great.  Thanks!

PR #255, opened 2020-03-30
Memento v1.0

ccoffrin:
If there are no objections, please merge and tag.


ccoffrin:
note, Memento v0.13 was tested locally and passed.

PR #258, opened 2020-04-13
Updates for PowerModels v0.16

ccoffrin:
@pseudocubic and @sanderclaeys these are the revisions that are required for PM v0.16.  Given that you will need to bump your minor version to v0.9, you might want to wait a week to see if PM v0.17 comes out (a function rename update) and put both updates into one release.

Also, in this PR I did not update to the convention of `const _PM = PowerModels` and `const _IM = InfrastructureModels`, but we might consider doing that update in the next release as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## staged
 
+- nothing
+
+## v0.9.0
+
+- Update to PowerModels v0.16
 - Add support for Memento v0.13, v1.0
 
 ## v0.8.1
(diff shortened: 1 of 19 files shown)


sanderclaeys:
Perhaps this is a good moment to raise an issue I'm having with how PowerModels generates the core ref object. 

Once we have variable size branches, PowerModels will try to generate buspairs using the same number of conductors for all of them (as set by `:conductors`), which will break on the branches that do not have that size. 

I can hack around that by setting ':conductors' to 1, so that PM only attempts to access the first element of the branches. The buspairs are not used anywhere anymore, so for me that does not matter much. However, it would be nice to have a cleaner solution for that in the long run.

ccoffrin:
Concerning the buspairs code, once have variable size branches we can update that code to be compatible.  I suppose at the moment you don't use buspairs becouse you are not yet considering parallel branches or transformers.  Is that a reasonable use case in the future?  If so, buspairs will become useful.

pseudocubic:
I ported these changes into the data-model branch, which we are preparing for release soon

Thanks @ccoffrin !

PR #261, opened 2020-04-18
4w patch

sanderclaeys:
minor bug fixing to get 4w prototype to work


pseudocubic:
it looks like something weird happened with the git history, I will try to rework it so that your changes get added correctly. Once I think I've fixed it, I'll let you know and you can double check it still contains the fixes you need and you can merge it

PR #262, opened 2020-04-18
fix broken SDP relaxations

frederikgeth:
SDPUBFPowerModel, SDPLCMXUBFPowerModel, SOCNLPUBFPowerModel and SOCConicUBFPowerModel are broken on master. I added unit tests to catch regressions. The tests do take a little time. We should revisit and add more tests once there is a more manageable way to define case study OPF extensions (bounds).


frederikgeth:
`pf_bf` was also untested, so I added tests for that. We'll need to revisit sometime though, as in the W variable space, you end up with redundant constraints (and too few degrees of freedom) when you have both `constraint_mc_theta_ref` and `constraint_mc_voltage_magnitude_setpoint` on the reference bus.

ready for review/merge if travis finishes successfully

test/pf_bf.jl
```diff
@@ -0,0 +1,120 @@
...
+        delete!(data["branch"], "3")
+        delete!(data["bus"], "4")
+        data["gen"]["1"]["gen_bus"] = 1
+
+        for (n, branch) in (data["branch"])
+                branch["rate_a"] = [10.0, 10.0, 10.0]
+        end
+
+        return data
+    end
```
pseudocubic, line 81: I don't think you need to define this as a function, just put it here in the outer test set context and the sub-testsets should all be able to use `data`.

It doesn't look like `data` gets altered below in any of the tests, so it would be better to avoid multiple parses of the same file

src/prob/pf_bf.jl
```diff
@@ -13,23 +13,38 @@ end
 ""
 function build_mc_pf_bf(pm::_PMs.AbstractPowerModel)
     # Variables
-    variable_mc_voltage(pm; bounded=false)
+    variable_mc_voltage(pm; bounded=true) #TODO should be false
```
pseudocubic, line 16: Is this issue still outstanding for this PR, or do you plan to address it later? What is the reason you have to set `bounded=true` here?

src/prob/pf_bf.jl
```diff
@@ -13,23 +13,38 @@ end
...
     # Constraints
     constraint_mc_model_current(pm)
 
     for (i,bus) in _PMs.ref(pm, :ref_buses)
-        constraint_mc_theta_ref(pm, i)
+        if typeof(pm)<:LPUBFDiagPowerModel
+            #do nothing, adding theta_ref would lead to redundant constraints
+        else
+            constraint_mc_theta_ref(pm, i)
+        end
```
pseudocubic, line 31: I would recommend condensing this into a single `if` rather than `if`-`else`, e.g.

```julia
if !(typeof(pm)<:LPUBFDiagPowerModel)
    constraint_mc_theta_ref(pm, i)
end
```

src/form/bf_mx.jl
```diff
@@ -9,6 +9,15 @@ end
...
     variable_mc_voltage_prod_hermitian(pm; kwargs...)
+    #enforce PSDness of leaf nodes that are not captured otherwise
+    for nw in _PMs.nw_ids(pm)
+        allbuses = Set(_PMs.ids(pm, nw, :bus))
+        startingbuses = Set(i for (l,i,j)  in _PMs.ref(pm, nw, :arcs_from))
+        leafnodes = setdiff(allbuses, startingbuses)
+        for i in leafnodes
+            constraint_mc_voltage_psd(pm, nw, i)
+        end
+    end
```
pseudocubic, line 20: I think the loop over `nw` should happen in the problem definition, should it not?

src/form/bf_mx.jl
```diff
@@ -64,6 +85,9 @@ function variable_mc_branch_series_current_prod_hermitian(pm::AbstractUBFModels;
             set_lower_bound_diag_to_zero=true, name="CC", prefix="$nw")
     end
 
+    for (id, L) in Lr
+        JuMP.set_start_value.(LinearAlgebra.diag(Lr[id]), 0.01)
+    end
```
pseudocubic, line 90: Is there a reason you use `Lr[id]` instead of `L` from the loop definition?

src/form/bf_mx.jl
```diff
@@ -28,7 +37,19 @@ function variable_mc_voltage_prod_hermitian(pm::AbstractUBFModels; n_cond::Int=3
         (Wr,Wi) = variable_mx_hermitian(pm.model, bus_ids, n_cond;
             set_lower_bound_diag_to_zero=true, name="W", prefix="$nw")
     end
-
+    v_start = exp.((im*2*pi/3).*[0; -1; 1]) #TODO this should be made more generic eventually
+    W_start = v_start*v_start'
+    for (id,_) in Wr
+        for i in 1:3
+            for j in 1:i
+                JuMP.set_start_value(Wr[id][i,j], real.(W_start)[i,j])
```
pseudocubic, line 45: Is there a reason to use `Wr[id]` rather than to include it in the loop definition?

src/form/bf_mx.jl
```diff
@@ -86,7 +110,11 @@ function variable_mc_branch_flow(pm::AbstractUBFModels; n_cond::Int=3, nw::Int=p
             bus_fr = _PMs.ref(pm, nw, :bus, branch["f_bus"])
             bus_to = _PMs.ref(pm, nw, :bus, branch["t_bus"])
 
-            smax_fr, smax_to = _calc_branch_power_ub_frto(branch, bus_fr, bus_to)
+            _calc_branch_power_max
```
pseudocubic, line 113: This seems to be dangling?

src/form/bf_mx.jl
```diff
@@ -210,34 +238,48 @@ function variable_mc_generation_power(pm::SDPUBFKCLMXModel; nw::Int=pm.cnw, boun
...
-    bound = Dict{eltype(gen_ids), Array{Real,2}}()
+    Pg_min = Dict{eltype(gen_ids), Array{Real,2}}()
+    Pg_max = Dict{eltype(gen_ids), Array{Real,2}}()
+    Qg_min = Dict{eltype(gen_ids), Array{Real,2}}()
+    Qg_max = Dict{eltype(gen_ids), Array{Real,2}}()
+
     for (id, gen) in _PMs.ref(pm, nw, :gen)
         bus = _PMs.ref(pm, nw, :bus, gen["gen_bus"])
-        vmax = bus["vmax"]
+        vmax = get(bus, "vmax", fill(Inf, 3))
```
pseudocubic, line 248: We probably don't want to hard-code 3 conductors in here, maybe refer to `pm.data["conductors"]`? That way it will be easier to find it later

src/form/bf_mx.jl
```diff
@@ -28,7 +37,19 @@ function variable_mc_voltage_prod_hermitian(pm::AbstractUBFModels; n_cond::Int=3
         (Wr,Wi) = variable_mx_hermitian(pm.model, bus_ids, n_cond;
             set_lower_bound_diag_to_zero=true, name="W", prefix="$nw")
     end
-
+    v_start = exp.((im*2*pi/3).*[0; -1; 1]) #TODO this should be made more generic eventually
+    W_start = v_start*v_start'
+    for (id,_) in Wr
+        for i in 1:3
```
pseudocubic, line 43: I think we don't want to hard-code 3 conductors in, maybe use the `size` of the matrix?

src/form/bf_mx.jl
```diff
@@ -210,34 +238,48 @@ function variable_mc_generation_power(pm::SDPUBFKCLMXModel; nw::Int=pm.cnw, boun
...
+        S_bound = _mat_mult_rm_nan(vmax, cmax')
+
+        Pg_min[id] = Qg_min[id] = -S_bound
+        Pg_max[id] = Qg_max[id] =  S_bound
+
+        pmin = get(gen, "pmin", fill(-Inf, 3))
+        pmax = get(gen, "pmax", fill( Inf, 3))
+        qmin = get(gen, "qmin", fill(-Inf, 3))
+        qmax = get(gen, "qmax", fill( Inf, 3))
+        for c in 1:3
```
pseudocubic, line 259: I think it would be better to use `conductor_ids` here

src/form/bf_mx.jl
```diff
@@ -329,45 +371,45 @@ function variable_mc_load(pm::SDPUBFKCLMXModel; nw=pm.cnw)
 end
 
 
+
```
pseudocubic, line 374: extra blank line

src/form/bf_mx.jl
```diff
@@ -785,33 +817,47 @@ S = U.I' = U.(Y.U)' = U.U'.Y' = W.Y'
...
-    # ignore dc for now
-    #TODO add DC in matrix version?
-    ncnds = size(Wr)[1]
+    Pg = get(_PMs.var(pm, n), :Pg_bus, Dict()); _PMs._check_var_keys(Pg, bus_gens, "active power", "generator")
+    Qg = get(_PMs.var(pm, n), :Qg_bus, Dict()); _PMs._check_var_keys(Qg, bus_gens, "reactive power", "generator")
+    Pd = get(_PMs.var(pm, n), :Pd_bus, Dict()); _PMs._check_var_keys(Pd, bus_loads, "active power", "load")
+    Qd = get(_PMs.var(pm, n), :Qd_bus, Dict()); _PMs._check_var_keys(Qd, bus_loads, "reactive power", "load")
+
+    cnds = _PMs.conductor_ids(pm; nw=n)
+    ncnds = length(cnds)
```
pseudocubic, line 832: Do these get used somewhere below that I don't see?

pseudocubic:
I implemented these fixes into the data-model branch, so they will appear in the next version

thanks for the contribution @frederikgeth !

PR #279, opened 2020-05-18
ADD: Store references for power balance constraints.

tasseff:
These additions store references of `:lam_kcl_r` and `:lam_kcl_i` into `con`.


pseudocubic:
Thanks @tasseff !

PR #287, opened 2020-05-26
REF: extract kron reduction, phase projection from eng2math

pseudocubic:
This PR refactors the kron reduction and padding transformations from within the individual eng2math conversion functions into their own functions inside of `src/data_models/transformations.jl` so that they can be applied to the engineering data model separately. This is in preparation for the addition of explicit 4-wire support to be added in v0.10.0.

In order to maintain backwards compatibility for the remainder of v0.9, the kron reduction and padding transformations are automatically applied during the call of the `eng2math` functions, and therefore during `transform_data_model`. This automatic transformation is expected to be removed at a future date. Furthermore, the user will not see the engineering data model after these transformations have been applied (unless they are invoked separately by the user) in all of the normal workflows. As can be seen in the file changes, unit test updates were very minor.

This PR adds two functions, `apply_kron_reduction!` and `apply_phase_projection!`, the calling of which will add `"is_kron_reduced"=>true` and `"is_projected"=>true`, respectively, to the top-level of the engineering data structure (and will propagate down to the mathematical data structure). `apply_kron_reduction!` performs two actions, (1) removes the neutral conductors of lines and compensates them through a kron reduction of the impedance, (2) removes all neutral terminals from the problem. `is_kron_reduced` will therefore indicate a departure from the engineering data model specification, where grounded neutrals are expected for certain objects, because 

`apply_phase_projection!` projects all components less than the total number of conductors over the whole network to that number of phases, _e.g._ if the network is overall three-phases, but there are some one and two phase components, the phases that are not connected on those components will be padded so that every component is three phases.

No documentation changes were required at this stage.

### DIFF
--- a/src/data_model/eng2math.jl
+++ b/src/data_model/eng2math.jl
@@ -64,14 +64,27 @@ end
 
 
 "base function for converting engineering model to mathematical model"
-function _map_eng2math(data_eng; kron_reduced::Bool=true)
+function _map_eng2math(data_eng::Dict{String,<:Any}; kron_reduced::Bool=true)
     @assert get(data_eng, "data_model", MATHEMATICAL) == ENGINEERING
 
+    # TODO remove kron reduction from eng2math in v0.10 (breaking)
+    _data_eng = deepcopy(data_eng)
+    if kron_reduced && !get(data_eng, "is_kron_reduced", false)
+        apply_kron_reduction!(_data_eng)
+    end
+
+    # TODO remove padding from eng2math in v0.10 (breaking)
+    if !get(data_eng, "is_padded", false)
+        apply_property_padding!(_data_eng)
+    end
+
     data_math = Dict{String,Any}(
-        "name" => get(data_eng, "name", ""),
-        "per_unit" => get(data_eng, "per_unit", false),
+        "name" => get(_data_eng, "name", ""),
+        "per_unit" => get(_data_eng, "per_unit", false),
         "data_model" => MATHEMATICAL,
-        "settings" => deepcopy(data_eng["settings"]),
+        "is_padded" => get(_data_eng, "is_padded", false),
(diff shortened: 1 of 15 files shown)


sanderclaeys:
I like the 'projection' terminology more, that's a good idea. I will review the PR by the end of the week and get back to you on the function name once I saw the code.

src/data_model/eng2math.jl
```diff
@@ -731,25 +620,28 @@ function _map_eng2math_voltage_source!(data_math::Dict{String,<:Any}, data_eng::
...
         if !all(isapprox.(get(eng_obj, "rs", zeros(1, 1)), 0)) && !all(isapprox.(get(eng_obj, "xs", zeros(1, 1)), 0))
+            f_bus = deepcopy(data_math["bus"]["$(math_obj["gen_bus"])"])
+
             bus_obj = Dict{String,Any}(
                 "bus_i" => length(data_math["bus"])+1,
                 "index" => length(data_math["bus"])+1,
-                "terminals" => collect(1:nconductors+1),
-                "grounded" => [fill(false, nconductors)..., true],
+                "terminals" => f_bus["terminals"],
+                "grounded" => f_bus["grounded"],
```
sanderclaeys, line 638: As we discussed in the previous PR, this is inconsistent with my mental model of what a voltage source represents. We can come back to this in v0.10 to decide what this should be, after discussing what it actually represents. In the meanwhile, if you decide to keep these changes, be aware this will break `PMD4W.jl` and you should use an older version of `PMD` with it for now.

pseudocubic: I have made some corrections to this section, and have confirmed that PMD4W.jl produces the same result with both PMD master and this branch on the Iowa case, the only change I had to make was to add the kwarg `project_phases=true` to `transform_data_model`

PR #289, opened 2020-05-27
FIX- branch flow linear model LPUBFDiagPowerModel

smithagopinath:
I have made constraints of the branch flow model conditional on an edge actually being present

Also added transformer linear constraints in bf.jl

Also edited errors in linear power loss constraints in bf_mx_lin.jl

Edited variable bounds in variable.jl


pseudocubic:
Thanks for the PR @smithagopinath !

Based on the Pull Request Checks I looks like there are a couple of items that you might want to address while we work on the rest of the code review. 

In particular, could you 

- add a line to the CHANGELOG, 
- update the Project.toml to remove CPLEX and Revise, and 
- add a unit test? 

In the case of this PR, the documentation check could be ignored, so I wouldn't worry that it keeps failing.

Project.toml
```diff
@@ -37,6 +35,3 @@ JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 Juniper = "2ddba703-00a4-53a7-87a5-e8b9971dde84"
 SCS = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
-
-[targets]
-test = ["Cbc", "Ipopt", "Juniper", "SCS", "Test"]
```
pseudocubic, line 40: Targets section got removed, needs to be re-added

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
 
+## LPUBF
+
+- Add constraint_mc_transformer_power_yy(pm::LPUBFDiagModel,..)
+- Add constraint_mc_transformer_power_dy(pm::LPUBFDiagModel,..)
+- Edited constraint_mc_power_losses(pm::LPUBFDiagModel..)
+- Edited constraint_constraint_mc_model_voltage_magnitude_difference(pm::LPUBFD..) to handle buses with not all 4 terminals active
+- Edited variable bounds in variable.jl to fix inconsistencies
+- Added unit tests
+
```
pseudocubic, line 11: Should be condensed to a single bullet in the `staged` section, e.g. 

`- Add wye-wye and delta-wye constraints to LPUBFDiagModel formulation (#289)`

src/core/variable.jl
```diff
@@ -119,8 +119,8 @@ function variable_mc_branch_power_real(pm::_PM.AbstractPowerModel; nw::Int=pm.cn
     if bounded
         for (l,i,j) in ref(pm, nw, :arcs)
             smax = _calc_branch_power_max(ref(pm, nw, :branch, l), ref(pm, nw, :bus, i))
-            set_upper_bound.(p[(l,i,j)],  smax)
-            set_lower_bound.(p[(l,i,j)], -smax)
+            set_upper_bound.(p[(l,i,j)],  1000000) #hacked bounds as _calc_branch_power_max involves multiplication of 0 with inf which is undf
+            set_lower_bound.(p[(l,i,j)], -1000000)
```
pseudocubic, line 123: I think the `undf` should be handled inside of the `set_lower_bound` and `set_upper_bound` function (which are located in `/src/core/data.jl`), instead of adding a hack here, which could cause unintended problems.

src/core/variable.jl
```diff
@@ -152,8 +152,8 @@ function variable_mc_branch_power_imaginary(pm::_PM.AbstractPowerModel; nw::Int=
     if bounded
         for (l,i,j) in ref(pm, nw, :arcs)
             smax = _calc_branch_power_max(ref(pm, nw, :branch, l), ref(pm, nw, :bus, i))
-            set_upper_bound.(q[(l,i,j)],  smax)
-            set_lower_bound.(q[(l,i,j)], -smax)
+            set_upper_bound.(q[(l,i,j)],  1000000) #hacked bounds as _calc_branch_power_max involves multiplication of 0 with inf which is undf
+            set_lower_bound.(q[(l,i,j)], -1000000)
```
pseudocubic, line 156: same comment as above

src/core/variable.jl
```diff
@@ -345,13 +345,25 @@ function variable_mc_bus_voltage_magnitude_sqr(pm::_PM.AbstractPowerModel; nw::I
...
+                end
+                if(vmax[c]==Inf)
+                    set_upper_bound.((w[i])[c], 10000.0)
+                else
+                    set_upper_bound.((w[i])[c], max(a^2, vmax[c]^2)) #this upper bound was wrong in the code
+                end
+                if(vmin[c]>0)
+                    set_lower_bound.((w[i])[c], vmin[c]^2)
+                end
+                end
```
pseudocubic, line 366: In v0.9 we changed things to a default of no voltage bounds (in reality, [0, Inf], in order that they still be physical), so if special voltage bounds are needed a transformation function should be used (see `/src/data_model/transformations.jl` for examples)

We should not introduce hacks into the code in this way, they will be easy to lose track of later

src/form/bf.jl
```diff
@@ -1,4 +1,4 @@
-"This is duplicated at PowerModelsDistribution level to correctly handle the indexing of the shunts."
+"This is duplicated at PMD level to correctly handle the indexing of the shunts."
```
pseudocubic, line 1: Unintended change?

src/form/bf.jl
```diff
@@ -46,10 +46,56 @@ end
...
+
+    w_fr = var(pm, nw, :w)[f_bus]
+    w_to = var(pm, nw, :w)[t_bus]
+
+    for p in 1:nph
+        JuMP.@constraint(pm.model, 1000.0*w_fr[p] == 1000.0*(pol*tm_scale*tm[p])^2*w_to[p])
+    end
+
+    JuMP.@constraint(pm.model, (p_fr + p_to)*1000.0 .== 0)
+    JuMP.@constraint(pm.model, (q_fr + q_to)*1000.0 .== 0)
```
pseudocubic, line 67: What are the reasons for the factors of 1e3? They seem like they can be eliminated cleanly.

src/form/bf.jl
```diff
@@ -46,10 +46,56 @@ end
 
 "nothing to do, this model is symmetric"
```
pseudocubic, line 47: needs documentation of mathematical formulation here

src/form/bf.jl
```diff
@@ -46,10 +46,56 @@ end
...
+    for p in 1:nph
+        JuMP.@constraint(pm.model, 1000.0*w_fr[p] == 1000.0*(pol*tm_scale*tm[p])^2*w_to[p])
+    end
+
+    JuMP.@constraint(pm.model, (p_fr + p_to)*1000.0 .== 0)
+    JuMP.@constraint(pm.model, (q_fr + q_to)*1000.0 .== 0)
 end
 
 
 "TODO"
```
pseudocubic, line 71: Needs documentation of mathematical formulation here

src/form/bf.jl
```diff
@@ -46,10 +46,56 @@ end
...
+    q_to = [var(pm, nw, :qt, t_idx)[p] for p in t_cnd]
+
+    w_fr = var(pm, nw, :w)[f_bus]
+    w_to = var(pm, nw, :w)[t_bus]
+
+    for p in 1:nph
+        # rotate by 1 to get 'previous' phase
+        # e.g., for nph=3: 1->3, 2->1, 3->2
+        q = (p-1+1)%nph+1
+	    JuMP.@constraint(pm.model, 3.0*1000.0*(w_fr[p] + w_fr[q]) == 2.0*1000.0*(pol*tm_scale*tm[p])^2*w_to[p])
```
pseudocubic, line 91: check out the `roll` function, I think it should do what you want here

src/form/bf.jl
```diff
@@ -46,10 +46,56 @@ end
...
+    end
+
+    for p in 1:nph
+            # rotate by nph-1 to get 'previous' phase
+            # e.g., for nph=3: 1->3, 2->1, 3->2
+        q = (p-1+nph-1)%nph+1
+	    JuMP.@constraint(pm.model, 2*p_fr[p]*1000.0 == 1000.0*(-(p_to[p]+p_to[q])+(q_to[q]-q_to[p])/sqrt(3.0)))
+        JuMP.@constraint(pm.model, 2*q_fr[p]*1000.0 == 1000.0*((p_to[p]-p_to[q])/sqrt(3.0)-(q_to[q]+q_to[p])))
```
pseudocubic, line 99: where do the factors of 1e3 come from, they seem like they can be cleanly eliminated?

test/data/opendss/case3_unbalanced_missingedge.dss
```diff
@@ -0,0 +1,36 @@
...
+New Load.L1 phases=1  loadbus.1.0   ( 0.4 3 sqrt / )   kW=9   kvar=3  model=1
+New Load.L2 phases=1  loadbus.2.0   ( 0.4 3 sqrt / )   kW=6   kvar=3  model=1
+New Load.L3 phases=1  Primary.3.0   ( 0.4 3 sqrt / )   kW=6   kvar=3  model=1
+
+Set voltagebases=[0.4]
+Set tolerance=0.000001
+set defaultbasefreq=50
+Calcvoltagebases
+
+Solve
```
pseudocubic, line 36: I wonder if instead of creating a new file if we could make these changes after parse of the original in the tests that you added, particularly since it is only called in one place

src/form/bf_mx_lin.jl
```diff
@@ -36,8 +36,24 @@ function constraint_mc_power_losses(pm::LPUBFDiagModel, n::Int, i, f_bus, t_bus,
...
+    cnds = conductor_ids(pm; nw=n)
+    ncnds = length(cnds)
+	
+     for c in 1:ncnds
+     if((vmax1[c]!=0) && (vmax2[c]!=0))
+       JuMP.@constraint(pm.model, (p_fr[c] + p_to[c])*1000.0 == (g_sh_fr[c,c]*w_fr[c] +  g_sh_to[c,c]*w_to[c])*1000.0)
+       JuMP.@constraint(pm.model, (q_fr[c] + q_to[c])*1000.0 == (-b_sh_fr[c,c]*w_fr[c] + -b_sh_to[c,c]*w_to[c])*1000.0)
+    else
+       b=JuMP.@constraint(pm.model, (p_fr[c] + p_to[c])*1000.0 == 0)
+       JuMP.@constraint(pm.model, (q_fr[c] + q_to[c])*1000.0 == 0)
```
pseudocubic, line 54: I don't think the factors 1e3 should be needed here, they are on both sides and can be divided out cleanly

src/form/bf_mx_lin.jl
```diff
@@ -49,35 +65,53 @@ function constraint_mc_model_voltage_magnitude_difference(pm::LPUBFDiagModel, n:
...
     MP = 2*(real(Gamma).*r + imag(Gamma).*x)
     MQ = 2*(real(Gamma).*x - imag(Gamma).*r)
 
-    JuMP.@constraint(pm.model, w_to .== w_fr - MP*p_s_fr - MQ*q_s_fr)
+    for c in 1:ncnds
+    if((vmax1[c]!=0) && (vmax2[c]!=0))
+      a=JuMP.@constraint(pm.model, w_to[c]*1000.0 == 1000.0*(w_fr[c] - (MP[c,1]*p_s_fr[1]+MP[c,2]*p_s_fr[2]+MP[c,3]*p_s_fr[3]) - (MQ[c,1]*q_s_fr[1]+MQ[c,2]*q_s_fr[2]+MQ[c,3]*q_s_fr[3])))
+    else
+      JuMP.@constraint(pm.model, p_s_fr[c]*1000.0==0)
+      JuMP.@constraint(pm.model, q_s_fr[c]*1000.0==0)
```
pseudocubic, line 91: same comment as above

src/form/bf_mx_lin.jl
```diff
@@ -49,35 +65,53 @@ function constraint_mc_model_voltage_magnitude_difference(pm::LPUBFDiagModel, n:
...
+    end
+   end
 
 end
 
 
+
```
pseudocubic, line 98: extra line

src/form/bf_mx_lin.jl
```diff
@@ -49,35 +65,53 @@ function constraint_mc_model_voltage_magnitude_difference(pm::LPUBFDiagModel, n:
...
 "balanced three-phase phasor"
 function constraint_mc_theta_ref(pm::LPUBFDiagModel, n::Int, i::Int, va_ref)
     ncnds = length(conductor_ids(pm))
     @assert(ncnds >= 2)
-
     w = var(pm, n, :w)[i]
-    JuMP.@constraint(pm.model, w[2:ncnds]   .== w[1])
+    #JuMP.@constraint(pm.model, w[2:ncnds]   .== w[1])
```
pseudocubic, line 104: inadvertent comment out?

src/form/bf_mx_lin.jl
```diff
@@ -49,35 +65,53 @@ function constraint_mc_model_voltage_magnitude_difference(pm::LPUBFDiagModel, n:
...
 
 
 ""
 function constraint_mc_load_power_balance(pm::LPUBFDiagModel, nw::Int, i, bus_arcs, bus_arcs_sw, bus_arcs_trans, bus_gens, bus_storage, bus_loads, bus_gs, bus_bs)
     w = var(pm, nw, :w, i)
+ bus = ref(pm, nw, :bus, i) 
+  vmax=bus["vmax"]
 
-    p = get(var(pm, nw), :p, Dict()); _PM._check_var_keys(p, bus_arcs, "active power", "branch")
+	p = get(var(pm, nw), :p, Dict()); _PM._check_var_keys(p, bus_arcs, "active power", "branch")
```
pseudocubic, line 114: inconsistent whitespace?

src/form/bf_mx_lin.jl
```diff
@@ -92,36 +126,45 @@ function constraint_mc_load_power_balance(pm::LPUBFDiagModel, nw::Int, i, bus_ar
     ps   = get(var(pm, nw),   :ps, Dict()); _PM._check_var_keys(ps, bus_storage, "active power", "storage")
     qs   = get(var(pm, nw),   :qs, Dict()); _PM._check_var_keys(qs, bus_storage, "reactive power", "storage")
 
+
```
pseudocubic, line 129: extra line

src/form/bf_mx_lin.jl
```diff
@@ -92,36 +126,45 @@ function constraint_mc_load_power_balance(pm::LPUBFDiagModel, nw::Int, i, bus_ar
...
+        sum((q[a])[c]*1000.0 for a in bus_arcs)
+        + sum((qsw[a_sw])[c]*1000.0 for a_sw in bus_arcs_sw)
+        + sum((qt[a_trans])[c]*1000.0 for a_trans in bus_arcs_trans)
+        ==
+        sum((qg[g])[c]*1000.0 for g in bus_gens)
+        - sum((qs[s])[c]*1000.0 for s in bus_storage)
+        - sum((qd[d])[c]*1000.0 for d in bus_loads)
+        + sum(bs[c]*w[c]*1000.0 for bs in values(bus_bs)))
+   end
+   end
```
pseudocubic, line 163: I'm not sure of the purpose behind this change, previously it was automatically cleanly over all phases, but now there is a loop over all phases? If the purpose is to eliminate phases which were padded/projected, and not in the original data, I think this isn't really the best way to accomplish that

test/transformer.jl
```diff
@@ -118,4 +118,21 @@
...
+        end
+        @testset "3w_dyy_1" begin
+            eng = parse_file("../test/data/opendss/ut_trans_3w_dyy_1.dss")
+            sol = run_ac_mc_opf(eng, LPUBFDiagPowerModel, ipopt_solver)
+            @test norm(solution["bus"]["3"]["w"]-[0.86095, 0.81344, 0.80480], Inf) <= 1E-4
+        end
+        @testset "3w_dyy_2" begin
+        eng = parse_file("../test/data/opendss/ut_trans_3w_dyy_2.dss")
+        sol = run_ac_mc_opf(eng, LPUBFDiagPowerModel, ipopt_solver)
+        @test norm(solution["bus"]["3"]["w"]-[0.87086, 0.83270, 0.83208], Inf) <= 1E-4
```
pseudocubic, line 135: What is the source of the voltages that are used as comparison in the tests here?

test/opf_bf.jl
```diff
@@ -33,6 +33,12 @@
             @test isapprox(sum(sol["solution"]["voltage_source"]["source"]["pg"] * sol["solution"]["settings"]["sbase"]), 0.0214812; atol=2e-3)
             @test isapprox(sum(sol["solution"]["voltage_source"]["source"]["qg"] * sol["solution"]["settings"]["sbase"]), 0.00927263; atol=2e-3)
         end
+        @testset "3-bus unbalanced lpubfdiag opf_bf with only two terminals on load bus" begin
+        pmd = parse_file("../test/data/opendss/case3_unbalanced_missingedge.dss")
+        sol = run_mc_opf(pmd, LPUBFDiagPowerModel, ipopt_solver)
+        @test sol["termination_status"] == LOCALLY_SOLVED
+        @test norm(solution["bus"]["loadbus"]["w"]-[0.92234, 0.95778, 0.0], Inf) <= 1E-4
```
pseudocubic, line 40: What is the origin of the voltages used for comparison in this test?

src/form/bf.jl
smithagopinath: This is similar to how transformers constraints are implemented in src/form/acp.jl, so I shall leave it as it is for now

test/data/opendss/case3_unbalanced_missingedge.dss
smithagopinath: Yes, but it would be a little hard for me to do as I am not super familiar with both opendss and with the data model

src/form/bf_mx_lin.jl
smithagopinath: I am not fully sure that the constraint is correct., I think the voltage angles are meant to be equal as the name of the function suggests (constraint_mc_theta_ref), not the voltage magnitudes

src/form/bf.jl
smithagopinath: I have some writeup in latex, in overleaf,  not sure how to link it to the comments?

src/form/bf.jl
```diff
@@ -44,12 +44,54 @@ function variable_mc_bus_voltage_on_off(pm::LPUBFDiagModel; kwargs...)
...
+    q_to = [var(pm, nw, :qt, t_idx)[p] for p in t_cnd]
+
+    w_fr = var(pm, nw, :w)[f_bus]
+    w_to = var(pm, nw, :w)[t_bus]
+
+    for p in 1:nph
+        # rotate by 1 to get 'previous' phase
+        # e.g., for nph=3: 1->3, 2->1, 3->2
+        q = (p-1+1)%nph+1
+	    JuMP.@constraint(pm.model, 3.0*(w_fr[p] + w_fr[q]) == 2.0*(pol*tm_scale*tm[p])^2*w_to[p])
```
sanderclaeys, line 87: Do you have a derivation for this equation somewhere? I did it on paper a got a expression.

Project.toml
```diff
@@ -37,4 +37,4 @@ SCS = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Cbc", "Ipopt", "Juniper", "SCS", "Test"]
+test = ["Cbc", "Ipopt", "Juniper", "SCS", "Test"]
```
pseudocubic, line 40: There should be an extra line at the end of files

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # PowerModelsDistribution.jl Change Log
 
+## LPUBF
+
+- Add wye-wye and delta-wye constraints to LPUBFDiagModel formulation 
+
```
pseudocubic, line 6: could you remove the LPUBF section and move the bullet to be with the ones under staged?

src/form/bf_mx_lin.jl
pseudocubic: I think this is part of a discussion we have been having about reference bus constraints needing magnitudes to be fixed too, not just angles. I agree that this constraint is probably not in the right place though just based on the name of the function. Comments @frederikgeth or @sanderclaeys ?

src/form/bf_mx_lin.jl
```diff
@@ -49,35 +65,52 @@ function constraint_mc_model_voltage_magnitude_difference(pm::LPUBFDiagModel, n:
...
 
 
 ""
 function constraint_mc_load_power_balance(pm::LPUBFDiagModel, nw::Int, i, bus_arcs, bus_arcs_sw, bus_arcs_trans, bus_gens, bus_storage, bus_loads, bus_gs, bus_bs)
     w = var(pm, nw, :w, i)
+    bus = ref(pm, nw, :bus, i) 
+    vmax=bus["vmax"]
 
-    p = get(var(pm, nw), :p, Dict()); _PM._check_var_keys(p, bus_arcs, "active power", "branch")
+	  p = get(var(pm, nw), :p, Dict()); _PM._check_var_keys(p, bus_arcs, "active power", "branch")
```
pseudocubic, line 113: inconsistent indentation

src/prob/opf.jl
```diff
@@ -118,7 +118,7 @@ function build_mc_opf(pm::AbstractUBFModels)
     # Variables
     variable_mc_bus_voltage(pm)
     variable_mc_branch_current(pm)
-    variable_mc_branch_power(pm)
+    variable_mc_branch_power(pm; bounded=false)
```
pseudocubic, line 121: can you add a `# TODO: revert to bounded in v0.10` at the end of this line here?

src/form/bf.jl
pseudocubic: You can drop in the latex, see for example https://github.com/lanl-ansi/PowerModelsDistribution.jl/blob/master/src/form/acp.jl#L345-L369

pseudocubic:
@smithagopinath the travis tests are continuing to fail, do they pass locally for you?

sanderclaeys:
@smithagopinath I would suggest the following test to ensure that the implementation is correct. You rely on two assumptions in your linearization,
1. Line losses are negligible,
2. The voltage phasors are perfectly balanced.

Therefore, you can read in a test case, make sure that the voltage at the source bus is perfectly balanced (as is e.g. the case for `ut_trans_2w_dy_lead.dss`), and set all branch series impedances to a very small (non-zero) value. This will then ensure that the above conditions are met.

You can then compare the power drawn in each phase of the slack bus generator, between an exact formulation and your linearization. With the mentioned modifications, the results should be the same (right?).

smithagopinath:
> @smithagopinath I would suggest the following test to ensure that the implementation is correct. You rely on two assumptions in your linearization,
> 
> 1. Line losses are negligible,
> 2. The voltage phasors are perfectly balanced.
> 
> Therefore, you can read in a test case, make sure that the voltage at the source bus is perfectly balanced (as is e.g. the case for `ut_trans_2w_dy_lead.dss`), and set all branch series impedances to a very small (non-zero) value. This will then ensure that the above conditions are met.
> 
> You can then compare the power drawn in each phase of the slack bus generator, between an exact formulation and your linearization. With the mentioned modifications, the results should be the same (right?).

Hi Sander,

Thats a good idea. In assuming the system is perfectly balanced, should not the loads also then be the same for each phase? In that case, the exact model and the linear approximation should be nearly the same I think. Let me know what you think! 
Thanks!

sanderclaeys:
@smithagopinath I believe you actually want them to be unbalanced. It is not a requirement for the linearization to be correct (since you only use the factor that the voltages across phases have certain ratios, which does not depend on the power flows in any way when the series impedances are very small). Also, if you make the loading balanced, you might cover up for certain bugs, which only become apparent when the loading is unbalanced.

smithagopinath:
> @smithagopinath I believe you actually want them to be unbalanced. It is not a requirement for the linearization to be correct (since you only use the factor that the voltages across phases have certain ratios, which does not depend on the power flows in any way when the series impedances are very small). Also, if you make the loading balanced, you might cover up for certain bugs, which only become apparent when the loading is unbalanced.

Hi @sanderclaeys, I have added a test as you suggested in test/transformer.jl, Have also request David to check the test once. Thanks!

src/form/bf_mx_lin.jl
frederikgeth: ideally we support these cases:
1. a bus where both the voltage magnitude and angles are fixed
2. a bus where the voltage angles are fixed, the voltage magnitudes are equal between the phases, but not set to a specific value

Either condition is sufficient to guarantee that a formulation in the lifted variable space is meaningful. I'm happy with supporting just case 1 though. We'll need to change the implementation of constraint_mc_theta_ref for the SDP relaxation as well though.

src/core/variable.jl
```diff
@@ -1014,4 +1017,4 @@ function variable_mc_gen_power_setpoint_imaginary_on_off(pm::_PM.AbstractPowerMo
     end
 
     report && _IM.sol_component_value(pm, nw, :gen, :qg, ids(pm, nw, :gen), qg)
-end
+end
```
pseudocubic, line 1020: there should be an additional newline at the end of this file

Project.toml
pseudocubic: Reminder that this issue is still outstanding

src/form/bf_mx_lin.jl
pseudocubic: @smithagopinath I think this should be changed back to being uncommented, and I will make a Github Issue that this should be addressed.

src/form/bf_mx_lin.jl
```diff
@@ -95,33 +128,41 @@ function constraint_mc_load_power_balance(pm::LPUBFDiagModel, nw::Int, i, bus_ar
...
-        + sum(qt[a_trans] for a_trans in bus_arcs_trans)
-        .==
-        sum(qg[g] for g in bus_gens)
-        - sum(qs[s] for s in bus_storage)
-        - sum(qd[d] for d in bus_loads)
-        + sum(bs.*w for bs in values(bus_bs))
-    )
-
-    con(pm, nw, :lam_kcl_r)[i] = cstr_p
-    con(pm, nw, :lam_kcl_i)[i] = cstr_q
```
pseudocubic, line 121: these need to be added back, you will need to change `cstr_p` and `cstr_q` instead the loop over the conductors to another variable e.g. `cp` and `cq`, and add `push!(cstr_p, cp)`, `push!(cstr_q, cq)` to after their creation for this to work correctly

pseudocubic: During the last update you got almost everything, but indavertantly left out putting these back
```julia
con(pm, nw, :lam_kcl_r)[i] = cstr_p
con(pm, nw, :lam_kcl_i)[i] = cstr_q
```

test/transformer.jl
```diff
@@ -140,4 +140,28 @@
...
+            eng = parse_file("../test/data/opendss/ut_trans_3w_dyy_2.dss")
+            sol = run_mc_opf(eng, LPUBFDiagPowerModel, ipopt_solver)
+            @test norm(sol["solution"]["bus"]["3"]["w"]-[0.87086, 0.83270, 0.83208], Inf) <= 1E-4
+        end
+        @testset "2w_dy_lead_small_series_impedance" begin
+            eng = parse_file("../test/data/opendss/ut_trans_2w_dy_lead_small_series_impedance.dss", data_model=MATHEMATICAL)
+            sola = run_mc_opf(eng, LPUBFDiagPowerModel, ipopt_solver)
+            solb = run_mc_opf(eng, ACPPowerModel, ipopt_solver)
+            @test norm(sola["solution"]["bus"]["1"]["w"]-solb["solution"]["bus"]["1"]["vm"].^2, Inf) <= 1.2E-3
+            @test norm(sola["solution"]["branch"]["1"]["pf"]-solb["solution"]["branch"]["1"]["pf"], Inf) <= 1E-3
```
sanderclaeys, line 164: @smithagopinath 

This looks great, thanks for adding these tests.

Could you change this to a test on the relative error instead of the absolute error? Depending on the `sbase`, this absolute error can be acceptable or not, whilst for the relative error, `1E-3` should always be good enough.

PR #293, opened 2020-06-11
FIX: typo on line 548 of form/ivr.jl

MartaVanin:
`sol(pm, nw, :gen, id)[:cig_bus] = var(pm, nw, :crg_bus, id)` corrected to `sol(pm, nw, :gen, id)[:cig_bus] = var(pm, nw, :cig_bus, id)`

Should I make an entry in the changelog for this very small fix?


pseudocubic:
@MartaVanin thanks for the PR!

Yes, if you could put a brief entry in the changelog that would be great. The other checks can be safely ignored.

MartaVanin:
changelog now updated

PR #294, opened 2020-06-18
ADD: possibility to give vm_start to variable_mc_voltage in ivr and acr

MartaVanin:
I am working on a package extension of PowerModelsDistribution where it could be interesting to initialize vr and vi in the acr and ivr formulations based on both a given "va_start" and "vm_start".
Currently, you can only pass a "va_start" value, while vm is defaulted to 1.
It would be convenient to have it in PMD already, if you approve.


src/form/ivr.jl
```diff
@@ -112,13 +112,13 @@ function variable_mc_bus_voltage(pm::_PM.AbstractIVRModel; nw=pm.cnw, bounded::B
...
-    vm = 1
+    vm = haskey(ref(pm, nw, :bus, id), "vm_start") ? busref["vm_start"] : fill(1.0, ncnd)
     for id in ids(pm, nw, :bus)
         busref = ref(pm, nw, :bus, id)
         if !haskey(busref, "va_start")
             for c in 1:ncnd
-                vr = vm*cos(theta[c])
-                vi = vm*sin(theta[c])
+                vr = vm.*cos(theta[c])
+                vi = vm.*sin(theta[c])
```
pseudocubic, line 121: I think you need to put `[c]` on `vm`, otherwise when you add the test it will error out because you're trying to set the start value as an array

src/form/acr.jl
```diff
@@ -11,13 +11,13 @@ function variable_mc_bus_voltage(pm::_PM.AbstractACRModel; nw=pm.cnw, bounded::B
...
-    vm = 1
+    vm = haskey(ref(pm, nw, :bus, id), "vm_start") ? busref["vm_start"] : fill(1.0, ncnd)
     for id in ids(pm, nw, :bus)
         busref = ref(pm, nw, :bus, id)
         if !haskey(busref, "va_start")
             for c in 1:ncnd
-                vr = vm*cos(theta[c])
-                vi = vm*sin(theta[c])
+                vr = vm.*cos(theta[c])
+                vi = vm.*sin(theta[c])
```
pseudocubic, line 20: I think you need to put `[c]` on `vm`, otherwise when you add the test it will error out because you're trying to set the start value as an array

pseudocubic:
@MartaVanin thanks for the PR, it looks like an interesting feature! 

Could you add some simple unit tests? 

Also, once you resolve the conflicts the automatic tests should run.

MartaVanin:
I am not sure why my test fails on travis, it works for me locally.
And actually I work on windows, not on mac, while the only instances in which it passes on trevis are mac. Could it have to do with how Ipopt is implemented remotely?

MartaVanin:
> Could you add some simple unit tests?
> 
> Also, once you resolve the conflicts the automatic tests should run.

@pseudocubic , I am not sure why the first two tests fail, as I did add a line to the changelog and a small unit test. 
I did not add documentation, because this is just a minor extension of something that was already there.
Would you like me to add something else? The unit test is very basic, but I am not sure I should extend this further.

pseudocubic:
> > Could you add some simple unit tests?
> > Also, once you resolve the conflicts the automatic tests should run.
> 
> @pseudocubic , I am not sure why the first two tests fail, as I did add a line to the changelog and a small unit test.
> I did not add documentation, because this is just a minor extension of something that was already there.
> Would you like me to add something else? The unit test is very basic, but I am not sure I should extend this further.

There's a bug in those checks, where it seems like the whole diff isn't being used maybe, so I wouldn't worry about it. I've confirmed that you've added something to the changelog and unit tests. 

It might be useful to add something very brief to the documentation, I don't think va_start, vm_start is documented at all. Maybe add a couple of sentences to `docs/src/quickguide.md` in like a new section at the end called something like Warm starting?

As for the unit tests, I guess my question there is does this return a result in less iterations or something? We need to have a way to check if vm_start has an actual effect on the result compared to running without vm_start

MartaVanin:
> It might be useful to add something very brief to the documentation, I don't think va_start, vm_start is documented at all. Maybe add a couple of sentences to `docs/src/quickguide.md` in like a new section at the end called something like Warm starting?
> 
> As for the unit tests, I guess my question there is does this return a result in less iterations or something? We need to have a way to check if vm_start has an actual effect on the result compared to running without vm_start

Sounds good, I'll add some line in the quickguide.
It is indeed a question whether the start is convenient or not, if you choose a bad start you could in principle be worse off than without. But my idea would be that it is up to the user to verify that, if they provide a non-flat start. If no start is provided, the code defaults to a flat start.
In the unit test, I give the result of a power flow as starting value of a new identical power flow, so normally that should ease the solution process. But I am not sure how to verify that it results in fewer iterations: in the power flow solution dictionary, this is not returned, and to edit that (assuming JuMP allows to retrieve the #iterations ) I would have to edit the solution builder in infrastructuremodels.jl, I am afraid. So this might require a discussion with the developers of that package. In the solution dictionary information is given on the solver time, but I don't think it is a very reliable indicator..

pseudocubic:
> > As for the unit tests, I guess my question there is does this return a result in less iterations or something? We need to have a way to check if vm_start has an actual effect on the result compared to running without vm_start

> In the unit test, I give the result of a power flow as starting value of a new identical power flow, so normally that should ease the solution process. But I am not sure how to verify that it results in fewer iterations: in the power flow solution dictionary, this is not returned, and to edit that (assuming JuMP allows to retrieve the #iterations ) I would have to edit the solution builder in infrastructuremodels.jl, I am afraid. So this might require a discussion with the developers of that package. In the solution dictionary information is given on the solver time, but I don't think it is a very reliable indicator..

That is a good point, @ccoffrin do you have any ideas about how we might unit test a warm start?

MartaVanin:
@pseudocubic I have added the documentation part, so you can check if it is okay for you. I wondered if you had any further input on the unit tests

pseudocubic:
Thanks for adding the documentation, that will be really helpful for new users, and thanks for the PR!

MartaVanin:
Thank you for your work!

PR #295, opened 2020-06-18
FIX: Storage Model

pseudocubic:
This PR should fix several problems with the storage model being applied inconsistently across the different problem specifications and formulations.

Now, storage should be supported in 
- ACP
- ACR
- DCP
- NFA
- LPUBFDiag

and
- OPF
- PF
- MLD

For the MLD problem, the definition in PowerModelsRestoration was followed as a template, and for OPF, PF, the definition in PowerModels was used for confirmation

There were two bugs in the parsing of storage objects,
1) kwrated was being used instead of kwhrated from dss for the engery_ub value in the engineering model
2) units were being converted in two places, but should have only be converted in the unit conversion functions

There were some other bugs that were encountered during the development of this PR and fixed
- bug in `con` mapping in power balance constraints where an array was expected
- bug in `constraint_mc_shed_power_balance` for LPUBFModel where Gt, Bt were not defined
- bug in the status of virtual branches created from switches where the status of the switch was not taken into account, only the state
- bug in type enforcement of `map` argument in `transform_solution`
- bug in transformer parsing where `status` wasn't being included

CC @tasseff @frederikgeth


src/form/acr.jl
```diff
@@ -47,6 +121,31 @@ function constraint_mc_voltage_magnitude_bounds(pm::_PM.AbstractACRModel, n::Int
...
+    vi = var(pm, n, :vi, i)
+    z_voltage = var(pm, n, :z_voltage, i)
+
+    for c in conductor_ids(pm, n)
+        if isfinite(vmax[c])
+            JuMP.@constraint(pm.model, vr[c]^2 + vi[c]^2 <= vmax[c]^2*z_voltage)
+        end
+
+        if isfinite(vmin[c])
+            JuMP.@constraint(pm.model, vr[c]^2 + vi[c]^2 >= vmin[c]^2*z_voltage)
```
ccoffrin, line 144: Note, this is a non-convex constraint.  Take a note that we should look into ways to avoid it in the future.

src/form/acr.jl
```diff
@@ -47,6 +121,31 @@ function constraint_mc_voltage_magnitude_bounds(pm::_PM.AbstractACRModel, n::Int
...
+end
+
+
+"on/off bus voltage magnitude constraint"
+function constraint_mc_bus_voltage_magnitude_on_off(pm::_PM.AbstractACRModel, n::Int, i::Int, vmin, vmax)
+    vr = var(pm, n, :vr, i)
+    vi = var(pm, n, :vi, i)
+    z_voltage = var(pm, n, :z_voltage, i)
+
+    for c in conductor_ids(pm, n)
```
ccoffrin, line 138: In this section you could also add these constraints,
```
z_voltage*vr_lb[c] <= vr[c] <= z_voltage*vr_ub[c]
z_voltage*vi_lb[c] <= vi[c] <= z_voltage*vi_ub[c]
```

src/form/acr.jl
```diff
@@ -269,8 +368,78 @@ function constraint_mc_load_power_balance(pm::_PM.AbstractACRModel, nw::Int, i::
...
+
+    if _IM.report_duals(pm)
+        sol(pm, nw, :bus, i)[:lam_kcl_r] = cstr_p
+        sol(pm, nw, :bus, i)[:lam_kcl_i] = cstr_q
+    end
+end
+
+
+""
+function constraint_mc_shed_power_balance(pm::_PM.AbstractACRModel, nw::Int, i::Int, bus_arcs, bus_arcs_sw, bus_arcs_trans, bus_gens, bus_storage, bus_pd, bus_qd, bus_gs, bus_bs)
```
ccoffrin, line 382: In PMR this is called `constraint_power_balance_shed`, consider moving the location of `shed`.

src/form/acr.jl
```diff
@@ -562,3 +731,34 @@ function constraint_mc_transformer_power_dy(pm::_PM.AbstractACRModel, nw::Int, t
         )
     end
 end
+
+
+""
+function constraint_mc_storage_losses(pm::_PM.AbstractACRModel, i::Int; nw::Int=pm.cnw, kwargs...)
```
ccoffrin, line 737: The `constraint_storage_losses` in PowerModels is already generalized to the MC case.  So you I would put this implementation in PowerModels.

ccoffrin: I would also expect that the constraint template for `constraint_mc_storage_losses` dispatches to the PM's `constraint_storage_losses` in all cases.

src/prob/mld.jl
```diff
@@ -28,7 +28,8 @@ function build_mc_mld(pm::_PM.AbstractPowerModel)
     variable_mc_gen_indicator(pm; relax=true)
     variable_mc_gen_power_setpoint_on_off(pm)
 
-    variable_mc_storage_power_mi(pm; relax=true)
+    variable_mc_storage_indicator(pm, relax=true)
+    variable_mc_storage_power_mi_on_off(pm, relax=true)
```
ccoffrin, line 32: As discussed before, we can probably get away with the the non-on/off here.

src/form/acp.jl
```diff
@@ -114,8 +114,8 @@ function constraint_mc_slack_power_balance(pm::_PM.AbstractACPModel, nw::Int, i:
 
     end
 
-    con(pm, nw, :lam_kcl_r)[i] = cstr_p
-    con(pm, nw, :lam_kcl_i)[i] = cstr_q
+    con(pm, nw, :lam_kcl_r)[i] = isa(cstr_p, Array) ? cstr_p : [cstr_p]
+    con(pm, nw, :lam_kcl_i)[i] = isa(cstr_q, Array) ? cstr_q : [cstr_q]
```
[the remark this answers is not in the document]
pseudocubic: If the constraint is for some reason [0, 0, 0] .== [0, 0, 0], JuMP automatically makes it 0 .== 0, which makes it not an array, and these lines fail with a type error

src/form/acr.jl
pseudocubic: I will make a new Issue for this

src/form/acp.jl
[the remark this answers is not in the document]
ccoffrin: I just did this test,
```
using JuMP # v0.21.3
m = Model()
c = @constraint([0, 0, 0] .== [0, 0, 0])
length(c) # is 3
```
Maybe a recently fixed bug?

src/form/acr.jl
```diff
@@ -47,6 +121,31 @@ function constraint_mc_voltage_magnitude_bounds(pm::_PM.AbstractACRModel, n::Int
...
+
+
+"on/off bus voltage magnitude constraint"
+function constraint_mc_bus_voltage_magnitude_on_off(pm::_PM.AbstractACRModel, n::Int, i::Int, vmin, vmax)
+    vr = var(pm, n, :vr, i)
+    vi = var(pm, n, :vi, i)
+    z_voltage = var(pm, n, :z_voltage, i)
+
+    for c in conductor_ids(pm, n)
+        if isfinite(vmax[c])
```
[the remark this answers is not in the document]
pseudocubic: This is used this way because by default we don't have voltage bounds anymore in the engineering model, which, because PowerModels always expects them, we convert to vmin = 0 and vmax=Inf for each phase

src/form/acr.jl
pseudocubic: That makes sense that it should go into PowerModels instead. In the short term, in order to get to our releases on schedule would you suggest we attempt to quickly add this to PowerModels, or make an Issue here and in PowerModels, and merge this function as is?

src/prob/mld.jl
pseudocubic: I will make sure there's an Issue to move this functionality to PowerModelsRestoration and simplify the implementation here

src/form/acp.jl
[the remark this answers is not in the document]
pseudocubic: Hmm, you are right, I don't get the error with that simple example, so the problem must be deeper, because I have confirmed with a different network that I still get the error with the most recent version of JuMP. I will investigate further.

src/core/constraint_template.jl
```diff
@@ -441,8 +441,9 @@ end
 "voltage magnitude setpoint constraint"
 function constraint_mc_voltage_magnitude_only(pm::_PM.AbstractPowerModel, i::Int; nw::Int=pm.cnw, kwargs...)
     bus = ref(pm, nw, :bus, i)
-    vmref = bus["vm"] #Not sure why this is needed
-    constraint_mc_voltage_magnitude_only(pm, nw, i, vmref)
+    if haskey(bus, "vm")
```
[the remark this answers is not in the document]
pseudocubic: The reason for this is that in the engineering model we only specify vm if we expect the voltage magnitude to be set at that bus, and therefore the bus_type when converting to the mathematical model will only be =2 if vm exists, but PowerModels will change any bus_type to =2 if a generator is connected during one of the correct_network calls

Issue #269, opened 2020-04-29
Inverter objects


pseudocubic:
from @hakanergun:

My suggestion would be to have a separate inverter model which can connect "DC" sources such as PV and batteries with the AC grid, rather than having the inverters as part of the PV or storage systems. This would help to avoid having inverters double (thus double the losses), if one would like to connect PV and batteries on the DC side. That would also help to connect many other items interchangeably, I think.

See original discussion at https://github.com/sanderclaeys/pmd-data-model/issues/22

pseudocubic:
Copied from #310 

> I asked @bluejuniper and @jtabarez, who have experience modeling inverters in a PMD extension package, to weigh in on  this PR, and start a discussion about converters in PMD in general, including what are the use cases, and what implementation should look like.
> 
> What I'm hoping is that we can get an impression on what the broader use case for converters/inverters in distribution modeling as their own object, what we cannot adequately model currently without them, and how common they really are as a distribution-level object. Were we to add something like this to PMD, I think we want to be careful to ensure that it be supported at all levels of the code, and be generic enough in its implementation, particularly at the mathematical model level, to support most of the common use cases.
> 
> Whether we need explicit DC buses was also something that was brought up in some offline discussion.

CC @ccoffrin @hei06j

PR #310, opened 2020-08-06
ADD: split storage and solar components to storage/converter and solar/converter components.

hei06j:
## Goal
We split the storage and solar components in the mathematical model to storage/converter and solar/converter components and changed them back to the storage and solar in the engineering model afterwards. We chose the term 'converter' as an abstraction of 'inverter'. The work in this PR was prototyped together with @frederikgeth. 

This is to be used for further use cases such as EV/converter. For now we assume a 1:1 relationship for converter:storage and converter:pv. Supporting 1:many would require re-thinking of the engineering-level result dictionary structure.

We would like to start a discussion on the approach and the way it is implemented. If there is interest in this contribution, I am happy to write the documentation and unit tests. PV system modelling closer to the OpenDSS spec can be more easily implemented as well if there is interest. EVs would be added next. 

The changes are targeted to be non-breaking (to be confirmed by unit testing). 

## Major changes:
- In the openDSS parser, pref, ref, rs, XS, and efficiency_curve parameters are stored.
- In the common.jl file, extensions "ref_add_converter_storage!, ref_add_bus_converter!, ref_add_converter_solar!" are added.
- mld, opf, and pf formulations are adapted.
- introduces an empty variable `variable_converter_device_power`, i.e. `pdc` dict (naming suggestions welcome), to represent the input on the secondary side of the converter (e.g. in case of inverter, the dc side). It is initialized in `constraint_converter_storage_balance`/ `constraint_converter_solar_balance`, i.e. we store the expressions for the PV/storage subsytem output (`pref-pcurt` or `sd-sc`) there so that `constraint_mc_converter_losses` can pick them up. 
- Refactors variables for modeling storage subsystems into storage and converter variables (the subsystems are therefore not `_mc` anymore)
- Refactors functions "variable_mc_storage_power_mi" and "variable_mc_storage_power_mi_on_off" to "variable_storage_power_mi" and "variable_mc_converter_power_on_off"

- Adds functions:
   - variable_converter_device_power
   - variable_solar_curtailment_power_real (still need to re-add fixed output PV)
   - constraint_converter_storage_balance (KCL-like constraint on the 'dc' side)
   - constraint_converter_solar_balance
   - _safely_store_data_eng! to avoid overwriting engineering objects if they already exist
   - _map_math2eng_converter
   - _map_converter_to_devices (to enable backward compatibility)

- Adapts functions to converter:
  - constraint_mc_storage_thermal_limit -> constraint_mc_converter_thermal_limit
  - constraint_mc_storage_losses ->    constraint_mc_converter_losses
  - constraint_mc_storage_on_off -> constraint_mc_converter_on_off
  - variable_mc_storage_current -> variable_mc_converter_current
  - variable_mc_storage_power_control_imaginary -> variable_mc_converter_power_control_imaginary
  - variable_mc_storage_power_control_imaginary_on_off -> variable_mc_converter_power_control_imaginary_on_off
  - variable_mc_storage_power_real -> variable_mc_converter_power_real
  - variable_mc_storage_power_imaginary -> variable_mc_converter_power_imaginary
  - variable_mc_storage_indicator -> variable_mc_converter_indicator
  - variable_mc_storage_power_on_off -> variable_mc_converter_power_on_off
  - variable_mc_storage_power_real_on_off -> variable_mc_converter_power_real_on_off
  - variable_mc_storage_power_imaginary_on_off -> variable_mc_converter_power_imaginary_on_off


pseudocubic:
@hei06j thanks for the PR, this looks interesting, and it has spawned some interesting offline discussion internally about converters/inverters and at what level we should support them (in PMD, vs in an extension package).

I asked @bluejuniper and @jtabarez, who have experience modeling inverters in a PMD extension package, to weigh in on this PR, and start a discussion about converters in PMD in general, including what are the use cases, and what implementation should look like. 

What I'm hoping is that we can get an impression on what the broader use case for converters/inverters in distribution modeling as their own object, what we cannot adequately model currently without them, and how common they really are as a distribution-level object. Were we to add something like this to PMD, I think we want to be careful to ensure that it be supported at all levels of the code, and be generic enough in its implementation, particularly at the mathematical model level, to support most of the common use cases.

Whether we need explicit DC buses was also something that was brought up in some offline discussion.

I suggest we have this discussion in #269 

CC @ccoffrin

bluejuniper:
Hi @hei06j,  if I follow correctly this allows for a lossy transportation network model on the dc-side of inverters? I think that it would be immediately useful as it allows for PV inverters with battery storage on the dc bus. I don't quite follow what the model is from the variables & constraints listed in the PR but I'm assuming it is this one below, where each edge on the graph has a power capacity and efficiency curve associated with it?

Generalized Inverter Mesh

@pseudocubic, I do think that having dc buses would be helpful as that allows for explicitly representing converters as edges like the figure below, again assuming I'm following correctly. Bear in mind that I'm just thinking about dc buses in terms of nodes where power flows sum to zero, at least for modeling the internals of power electronics. 

Generalized Inverter

pseudocubic:
Because this PR is very behind the current master branch, and I think there probably needs to be a rethink of how to implement dc buses without changing existing storage constraints, i am going to close this PR. 

I really encourage anybody that is interested in us implementing dc buses to start a discussion here: https://github.com/lanl-ansi/PowerModelsDistribution.jl/discussions

Ultimately, I think the consensus is that we could find dc buses really valuable, but it hasn't been made clear to me how best to implement them yet.

Issue #319, opened 2020-12-16
Bug/enhancement: missing support for single-phase buses (v0.10.0)


MartaVanin:
Hello,
on a similar note of #318 , if I reduce three-phase buses to single-phase buses, a number of constraints (depending on the formulation) prevents their integration in  powerflow and opf problems:
```julia
using PowerModelsDistributions, Ipopt

ipopt_solver = optimizer_with_attributes(Ipopt.Optimizer, "print_level"=>0)

data = parse_file("/test/data/opendss/case3_unbalanced.dss"; data_model = MATHEMATICAL)
delete!(data["load"], "2")
delete!(data["load"], "3")
data["bus"]["3"]["terminals"] = [2]
data["bus"]["3"]["grounded"] = Bool[0.0]
data["branch"]["1"]["t_connections"] = [2]
data["branch"]["1"]["f_connections"] = [2]
#for instance, all of the following will fail:
run_mc_pf(data, ACRPowerModel, ipopt_solver)
run_mc_pf(data, IVRPowerModel, ipopt_solver)
run_mc_opf(data, ACPPowerModel, ipopt_solver)
run_mc_opf(data, ACRPowerModel, ipopt_solver)
run_mc_opf(data, IVRPowerModel, ipopt_solver)
```
I think the option of having single-phase buses/branches is quite valuable and should be supported.
One easy way to fix that could be to broadcast of the "failing constraints" over the intersection of the component's connections. For instance, `run_mc_opf(data, ACPPowerModel, ipopt_solver)` is "blocked" by `constraint_mc_voltage_angle_difference` at:
https://github.com/lanl-ansi/PowerModelsDistribution.jl/blob/master/src/form/shared.jl#L240-L248
What could solve it could be something like the following, if you find it acceptable:
```julia
function constraint_mc_voltage_angle_difference(pm::_PM.AbstractPolarModels, nw::Int, f_idx::Tuple{Int,Int,Int}, f_connections::Vector{Int}, t_connections::Vector{Int}, angmin::Vector{<:Real}, angmax::Vector{<:Real})
    i, f_bus, t_bus = f_idx

   common_connections = intersect(f_connections, t_connections)

    va_fr = [var(pm, nw, :va, f_bus)[fc] for fc in common_connections]
    va_to = [var(pm, nw, :va, t_bus)[tc] for tc in common_connections]

    JuMP.@constraint(pm.model, va_fr .- va_to .<= angmax)
[...]

PR #330, opened 2021-03-20
FIX: single-phase buses and branches

MartaVanin:
The scope of this pull request is to enable single-phase buses and branches, see #319  .
Currently, I did it for ivr, acr, acp and lindist3flow.  I think acp and ld3f look ok, whereas the ivr and acr workarounds are not very elegant (if loop that checks the dimension of impedance array + almost duplicated code).
This is because if impedance arrays are 1x1 and you try to multiply, e.g., a 1x1 `G` and a 1x1 `vr_to` like [here](https://github.com/lanl-ansi/PowerModelsDistribution.jl/blob/master/src/form/acr.jl#L557), julia will throw an error saying that the dimensions are not right for matrix multiplication.
On the other hand, if you pass the impedance values as scalars:
1) it makes the construction of the data dictionary less consistent, imho, 
2) to allow for 1x1 impedance elements I "downgraded" the method definitions in the interested constraints from `G::Matrix` to `G::Array`. If you would allow them to even be scalars, you would need a `Any` or a `Union`, but this also didn't seem very nice.
I suppose one workaround could be to edit something upstream, e.g., in `calc_branch_y` or  G, B in `constraint_mc_ohms_yt_to\from` in constraint_template.jl, but I wanted to check first what you think the best approach is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## staged
 
-- none
+- fix bug for acr, ivr, acp and lindist3flow forms, allowing elements with single phase terminals/connections
 
 ## v0.10.2
 
--- a/src/core/constraint_template.jl
+++ b/src/core/constraint_template.jl
@@ -100,6 +100,8 @@ function constraint_mc_ohms_yt_from(pm::_PM.AbstractPowerModel, i::Int; nw::Int=
     t_idx = (i, t_bus, f_bus)
 
     G, B = _PM.calc_branch_y(branch)
+    if isa(G, Float64) G = [G] end
+    if isa(B, Float64) B = [B] end
 
     constraint_mc_ohms_yt_from(pm, nw, f_bus, t_bus, f_idx, t_idx, branch["f_connections"], branch["t_connections"], G, B, branch["g_fr"], branch["b_fr"])
 end
@@ -114,6 +116,8 @@ function constraint_mc_ohms_yt_to(pm::_PM.AbstractPowerModel, i::Int; nw::Int=pm
     t_idx = (i, t_bus, f_bus)
 
     G, B = _PM.calc_branch_y(branch)
+    if isa(G, Float64) G = [G] end
+    if isa(B, Float64) B = [B] end
 
     constraint_mc_ohms_yt_to(pm, nw, f_bus, t_bus, f_idx, t_idx, branch["f_connections"], branch["t_connections"], G, B, branch["g_to"], branch["b_to"])
 end
--- a/src/core/constraint_template.jl
+++ b/src/core/constraint_template.jl
@@ -100,10 +100,8 @@ function constraint_mc_ohms_yt_from(pm::_PM.AbstractPowerModel, i::Int; nw::Int=
     t_idx = (i, t_bus, f_bus)
 
     G, B = _PM.calc_branch_y(branch)
-    if isa(G, Float64) G = [G] end
-    if isa(B, Float64) B = [B] end
 
-    constraint_mc_ohms_yt_from(pm, nw, f_bus, t_bus, f_idx, t_idx, branch["f_connections"], branch["t_connections"], G, B, branch["g_fr"], branch["b_fr"])
+    constraint_mc_ohms_yt_from(pm, nw, f_bus, t_bus, f_idx, t_idx, branch["f_connections"], branch["t_connections"], Array(G), Array(B), branch["g_fr"], branch["b_fr"])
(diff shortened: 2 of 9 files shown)


pseudocubic:
I think I'm still confused about the actual bug behind this.

i constructed the following test case in dss and ran `solve_mc_pf` and `solve_mc_opf` with ACP, ACR, IVR, and LPUBFDiag and didn't run into any errors:

```dss
Clear
New Circuit.3Bus_1p_example
~ basekv=0.4   pu=0.9959  MVAsc1=1e6  MVAsc3=1e6

New linecode.556MCM nphases=1 basefreq=50  ! ohms per 5 mile
~ rmatrix = ( 0.1000 )
~ xmatrix = ( 0.0583 )
~ cmatrix = ( 50.92958178940651 ) ! small capacitance

New linecode.4/0QUAD nphases=1 basefreq=50  ! ohms per 100ft
~ rmatrix = ( 0.1167 )
~ xmatrix = ( 0.0667 )
~ cmatrix = ( 50.92958178940651 )  ! small capacitance

New Line.OHLine  bus1=sourcebus.1  Primary.1  linecode = 556MCM   length=1  ! 5 mile line
New Line.Quad    Bus1=Primary.1  loadbus.1  linecode = 4/0QUAD  length=1   ! 100 ft

New Load.L1 phases=1  loadbus.1.0   ( 0.4 3 sqrt / )   kW=6   kvar=3  model=1

Set voltagebases=[0.4]
Set tolerance=0.000001
set defaultbasefreq=50
Calcvoltagebases

Solve

```

I suspect there is a datatype issue in the data model construction when building your single-phase model by hand rather than importing it from a dss file. Looking at the datatypes of the case3_reduced in your test example, it seems like they are incorrect. Looking back at #319, I actually wouldn't expect that to work, because all of the br_r,br_x,etc are the wrong size for the branch now, which would break the matrix multiplication bits, as you saw. To fix your example in #319, i would do this:

```julia
data = parse_file("test/data/opendss/case3_unbalanced.dss"; data_model = MATHEMATICAL)
delete!(data["load"], "2")
delete!(data["load"], "3")
data["bus"]["3"]["terminals"] = Vector{Int}([2])
data["bus"]["3"]["grounded"] = Vector{Bool}([0])

for (k,v) in data["branch"]["1"]
    if isa(v, Matrix)
        data["branch"]["1"][k] = reshape([v[2,2]], 1, 1)
    elseif isa(v, Vector)
        data["branch"]["1"][k] = [v[2]]
    end
end

result_pf = solve_mc_pf(data, ACRPowerModel, ipopt_solver)
result_opf = solve_mc_opf(data, ACRPowerModel, ipopt_solver)
```

This runs without errors in my dev environment (PowerModelsDistribution#master)

MartaVanin:
Hi @pseudocubic , thanks for looking into this. I tried your test fix; it also works for me and it passes the tests I originally created in "test/single_connections.jl" in this PR. So #319 looks like a false positive to me now, it seems that it was only due to the wrong datatypes in my test case.
I had a look at the size of the matrices for the ACP case, and I saw that in [constraint_mc_ohms_yt_from](https://github.com/lanl-ansi/PowerModelsDistribution.jl/blob/master/src/form/acp.jl#L463) (which in my original test case was one of the reasons for failure) all impedance-related matrices are 1x1. The "fr" variables (`vm_fr`, `va_fr`, `p_fr`, `q_fr`) are all size (1,), whereas `vm_to`, `va_to` are (3,), which is expected as that bus has still three terminals. But then these two variables are indexed as per the branch's active connections, so those pick the one wished element and all dimensions seem to be fine eventually.
If for some reason the connections of the branch in the data are not reduced to match the single-phase bus then it does fail, but that makes sense.
So I am a bit unsure on what to do with this PR actually, I was thinking to maybe just add the (fixed) single-phase connections test, as it might be handy, discard all other changes I made, and close issue 319.
Just to be sure I could first check the size of the other constraints first, but it seems that `_PM.calc_branch_y(branch)` + just reducing the terminals and connections in the data does the trick.

pseudocubic:
I think ultimately what I realized from your PR and Issue is that PMD is in need of some more data validation/correction functions, especially since there is a mix of types that all need to be carefully considered.  

I think for now we can just close the PR, because a few things that you mentioned are already in the pipeline for when we upgrade to add multi-infrastructure support (InfrastructureModels v0.6), and I can add in the two examples of a single phase problems when doing that.

MartaVanin:
OK, I'll just close this then.

PR #337, opened 2021-04-20
ADD: LineGeometry / LineSpacing / Wiredata

pseudocubic:
This adds support for LineGeometry, LineSpacing, Wiredata DSS objects.

It uses calculations from Kersting's "Distribution System Modeling and Analysis" to convert from LineGeometry and/or LineSpacing with Wiredata objects to series impedance and shunt admittance matrices. The implementation of this calculation can be found in the function `_calc_impedances_from_geometry` in `src/io/utils.jl`

Because we do not have LineGeometry objects in the ENGINEERING model, and no immediate plans to add an additional complex data object such as this, LineGeometry / Wiredata / Linespacing get converted to  series impedance and shunt admittance on the line on which they are specified in the DSS input file.

In its current state, this PR does not support CNData or TSData (underground cables). This is an additional conversion calculation that will need to be added.

In opendss, the earthmodel can be "deri", "fullcarson", or "carson". Deri is a curve fit to full carson which is supposed to be more accurate at higher frequencies. Carson is a simplified version of fullcarson that only uses the first terms of the expansion. Deri has not been implemented in this PR (fullcarson is used in place of deri)

This should also help with 4-wire support when it comes, because it seems more common for line properties of 4 conductor lines to be defined with linegeometry rather than linecodes.

Note that all units get converted to meters on parse, and therefore any equations need to also be in meters to be correct.

On LineGeometry objects, there is a feature similar to Transformers, where the "cond" can be set, and any properties set afterwards are assigned to that conductor. To support this, _add_property had to be refactored slightly to support keys and properties specific to the object type.

A bug in _parse_element_with_dtype was fixed, where values where left as Substrings instead of getting converted to String

# TODO

- [ ] Verify conversions are correct / debug if necessary
- [ ] Add unit tests
- [ ] Update changelog

### DIFF
--- a/src/io/dss/dss2eng.jl
+++ b/src/io/dss/dss2eng.jl
@@ -467,15 +467,31 @@ function _dss2eng_line!(data_eng::Dict{String,<:Any}, data_dss::Dict{String,<:An
             eng_obj["linecode"] = dss_obj["linecode"]
         end
 
-        if any(haskey(dss_obj, key) && _is_after_linecode(dss_obj["prop_order"], key) for key in ["r0", "r1", "rg", "rmatrix"]) || !haskey(dss_obj, "linecode")
+        from_geometry = haskey(dss_obj, "geometry") && _is_after_linecode(dss_obj["prop_order"], "geometry")
+        from_spacing = haskey(dss_obj, "wires") && _is_after_linecode(dss_obj["prop_order"], "wires")
+        if from_geometry || from_spacing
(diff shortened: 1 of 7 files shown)


sanderclaeys:
Looking forward to this feature!

I created a temporary workaround which uses OpenDSSDirect to obtain the impedance matrices for `Geometry` objects, and then dumps these in a temporary file (in memory) where the lines are changed to refer to these linecodes instead. I will link to it below, maybe it can be useful to you for debugging this.

https://www.dropbox.com/s/pbmu68haowsqwzh/parse_file_hybrid.jl?dl=0

PR #340, opened 2021-06-29
[ADD][4W-EXTENSION] IVR and ACR formulations

sanderclaeys:
**Work in progress, this will be updated throughout the day. I will indicate here when the description is finished**

# General discussion

# Changes per file
- `src/PowerModelsDistribution.jl`: added new files
- `src/core/base.jl`: 
    - simplified signatures (some keywords made it to master related to a feature that was never added)
    - added a method that allows a scalar bound to be assigned to a vector
- `src/core/data.jl`: added helper functions to get internal dimensions of a component based on component dictionary (concept described above)
- `src/core/objective.jl`: new methods for new formulations, to deal with generator indexing
- `src/core/types.jl`: added to type hierarchy, as described above
- `src/data_model/eng2math.jl`: 
    - missing bounds are added as `Inf` so the key is always present, consistent with behavior for `vmin`/`vmax` 
    - (except `vm_pair_lb`/`vm_pair_ub`, for which the equivalent is an empty vector)
- `src/data_model/math2eng.jl`: map power keys for transformers
- `src/data_model/units.jl`: map power keys for transformers
- `src/data_model/utils.jl`: related to changes in `eng2math.jl`
- `src/form/en_shared.jl`: methods typed on `ExplicitNeutralModels` or `RectangularVoltageExplicitNeutralModels`
- `src/form/en_ivr.jl`: methods typed on `AbstractExplicitNeutralIVRModel` or subtypes
- `src/form/en_acr.jl`: methods typed on `AbstractExplicitNeutralACRModel` or subtypes
- `src/prob/opf.jl`:
    - new methods for new formulations
    - note that for these both power and current bounds for lines and switches are imposed for both ACR and IVR
    - docstrings were added in the process for all public methods


sanderclaeys:
I started designing the unit tests for the next PR, and this is process is uncovering some bugs and typos in this PR. I will commit corrections for those by the end of the week.

sanderclaeys:
It seems like the switch bounds are called `current_rating` and `thermal_rating`, instead of `c_rating_a` and `rate_a`, as is the case for the switches. However, the data model transformation does seem to use `c_rating_a` and `rate_a`. For now, I corrected the data model transformation, but do check which names you would like to go with.

src/form/en_ivr.jl
```diff
@@ -0,0 +1,1763 @@
...
+For non-linear IVR models with explicit neutrals,
+imposes a bound on the magnitude of the total apparent power at both windings. 
+Expressions are created for the transformer power variables.
+
+```
+sum(pt_fr)^2 + sum(qt_fr)^2 <= sm_ub^2
+sum(pt_to)^2 + sum(qt_to)^2 <= sm_ub^2
+```
+"""
+function constraint_mc_transformer_power_rating(pm::AbstractNLExplicitNeutralIVRModel, nw::Int, id::Int, f_idx::Tuple, t_idx::Tuple, f_bus::Int, t_bus::Int, f_connections::Vector, t_connections::Vector, config::ConnConfig, sm_ub::Real; report::Bool=true)
```
pseudocubic, line 1084: should be `thermal_rating` to be consistent with other naming conventions

src/form/en_ivr.jl
```diff
@@ -0,0 +1,1763 @@
...
+		report::Bool=true,
+		kwargs...
+	)
+
+For quadratic IVR models with explicit neutrals,
+creates transformer power variables `:pt` and `:qt`
+"""
+function variable_mc_transformer_power(pm::AbstractQuadraticExplicitNeutralIVRModel; nw::Int=nw_id_default, bounded::Bool=true, report::Bool=true, kwargs...)
+    variable_mc_transformer_power_active(pm, nw=nw, bounded=bounded, report=report; kwargs...)
+    variable_mc_transformer_power_reactive(pm, nw=nw, bounded=bounded, report=report; kwargs...)
```
pseudocubic, line 912: should be `real` and `imaginary` instead of `active` and `reactive`, to be consistent with other naming conventions

src/form/en_ivr.jl
```diff
@@ -0,0 +1,1763 @@
...
+
+For quadratic IVR models with explicit neutrals,
+imposes a bound on the magnitude of the total apparent power at both windings.
+
+```
+sum(pt_fr)^2 + sum(qt_fr)^2 <= sm_ub^2
+sum(pt_to)^2 + sum(qt_to)^2 <= sm_ub^2
+```
+"""
+function constraint_mc_transformer_power_rating(pm::AbstractQuadraticExplicitNeutralIVRModel, nw::Int, id::Int, f_idx::Tuple, t_idx::Tuple, f_bus::Int, t_bus::Int, f_connections::Vector, t_connections::Vector, config::ConnConfig, sm_ub::Real; report::Bool=true)
```
pseudocubic, line 1161: Should be `thermal_rating` to be consistent with other naming conventions

src/form/en_ivr.jl
```diff
@@ -0,0 +1,1763 @@
...
+
+    if sm_ub < Inf
+        JuMP.@constraint(pm.model, sum(pt_fr)^2+sum(qt_fr)^2 <= sm_ub^2)
+        JuMP.@constraint(pm.model, sum(pt_to)^2+sum(qt_to)^2 <= sm_ub^2)
+    end
+    
+    # if report
+    #     sol(pm, nw, :transformer, id)[:sm_fr] = JuMP.@NLexpression(pm.model, sqrt(pt_fr^2+qt_fr^2))
+    #     sol(pm, nw, :transformer, id)[:sm_to] = JuMP.@NLexpression(pm.model, sqrt(pt_to^2+qt_to^2))
+    # end
```
pseudocubic, line 1208: any reason to leave this commented out?

src/form/en_ivr.jl
```diff
@@ -0,0 +1,1763 @@
...
+function constraint_mc_thermal_limit_to(pm::AbstractExplicitNeutralIVRModel, nw::Int, t_idx::Tuple{Int,Int,Int}, t_connections::Vector{Int}, rate_a::Vector{<:Real})
+    vr_to = [var(pm, nw, :vr, t_idx[1])[t] for t in t_connections]
+    vi_to = [var(pm, nw, :vi, t_idx[1])[t] for t in t_connections]
+    cr_to = var(pm, nw, :cr, t_idx)
+    ci_to = var(pm, nw, :ci, t_idx)
+
+    for idx in 1:length(rate_a)
+        if rate_a[idx]<Inf
+            psw_to_idx = JuMP.@NLexpression(pm.model,  vr_to[idx]*cr_to[idx] + vi_to[idx]*ci_to[idx])
+            qsw_to_idx = JuMP.@NLexpression(pm.model, -vr_to[idx]*ci_to[idx] + vi_to[idx]*cr_to[idx])
```
pseudocubic, line 1480: It looks like you wrote a generic `thermal_limit_to` function, so the `psw_to` and `qsw_to` make it a little confusing, I think change to just pt and qt

also, update the docstring to indicate that the constraint is generic to edge types

src/form/en_ivr.jl
```diff
@@ -0,0 +1,1763 @@
...
+function constraint_mc_thermal_limit_from(pm::AbstractExplicitNeutralIVRModel, nw::Int, f_idx::Tuple{Int,Int,Int}, f_connections::Vector{Int}, rate_a::Vector{<:Real})
+    vr_fr = [var(pm, nw, :vr, f_idx[1])[t] for t in f_connections]
+    vi_fr = [var(pm, nw, :vi, f_idx[1])[t] for t in f_connections]
+    cr_fr = var(pm, nw, :cr, f_idx)
+    ci_fr = var(pm, nw, :ci, f_idx)
+
+    for idx in 1:length(rate_a)
+        if rate_a[idx]<Inf
+            psw_fr_idx = JuMP.@NLexpression(pm.model,  vr_fr[idx]*cr_fr[idx] + vi_fr[idx]*ci_fr[idx])
+            qsw_fr_idx = JuMP.@NLexpression(pm.model, -vr_fr[idx]*ci_fr[idx] + vi_fr[idx]*cr_fr[idx])
```
pseudocubic, line 1451: It looks like you wrote a generic `thermal_limit_fr` function, so the `psw_fr` and `qsw_fr` make it a little confusing, I think change to just `pf` and `qf`

also, update the docstring to indicate that the constraint is generic to edge types

src/form/en_ivr.jl
```diff
@@ -0,0 +1,1763 @@
...
+
+For quadratic IVR models with explicit neutrals, 
+throw an error because this cannot be represented quadratically 
+without introducing explicit power variables.
+"""
+function constraint_mc_thermal_limit_from(pm::AbstractQuadraticExplicitNeutralIVRModel, nw::Int, f_idx::Tuple{Int,Int,Int}, f_connections::Vector{Int}, rate_a::Vector{<:Real})
+    if any(rate_a.<Inf)
+        error("""
+            A branch power bound cannot be represented quadratically in the default AbstractQuadraticExplicitNeutralIVRModel.
+            Either extend this quadratic formulation by including explicit branch power variables, or use AbstractNLExplicitNeutralIVRModel instead.""")
```
pseudocubic, line 1507: Should this be an error? I think it might be better to just be a warning and to ignore the thermal limit, which is more in-line with current behavior elsewhere in the package

src/form/en_ivr.jl
```diff
@@ -0,0 +1,1763 @@
...
+
+For quadratic IVR models with explicit neutrals, 
+throw an error because this cannot be represented quadratically 
+without introducing explicit power variables.
+"""
+function constraint_mc_thermal_limit_to(pm::AbstractQuadraticExplicitNeutralIVRModel, nw::Int, t_idx::Tuple{Int,Int,Int}, t_connections::Vector{Int}, rate_a::Vector{<:Real})
+    if any(rate_a.<Inf)
+        error("""
+            A branch power bound cannot be represented quadratically in the default AbstractQuadraticExplicitNeutralIVRModel.
+            Either extend this quadratic formulation by including explicit branch power variables, or use AbstractNLExplicitNeutralIVRModel instead.""")
```
pseudocubic, line 1529: Should this be an error? I think it might be better to just be a warning and to ignore the thermal limit, which is more in-line with current behavior elsewhere in the package

src/form/en_ivr.jl
```diff
@@ -0,0 +1,1763 @@
...
+
+For quadratic IVR models with explicit neutrals, 
+throw an error because this cannot be represented quadratically 
+without introducing explicit power variables.
+"""
+function constraint_mc_switch_thermal_limit(pm::AbstractQuadraticExplicitNeutralIVRModel, nw::Int, f_idx::Tuple{Int,Int,Int}, f_connections::Vector{Int}, rating::Vector{<:Real})
+    if any(rating.<Inf)
+        error("""
+            A switch power bound cannot be represented quadratically in the default AbstractQuadraticExplicitNeutralIVRModel.
+            Either extend this quadratic formulation by including explicit switch power variables, or use AbstractNLExplicitNeutralIVRModel instead.""")
```
pseudocubic, line 1761: Should this be an error? I think it might be better to just be a warning and to ignore the thermal limit, which is more in-line with current behavior elsewhere in the package

src/form/en_acr.jl
```diff
@@ -0,0 +1,839 @@
...
+    vr_fr   = var(pm, nw, :vr, f_idx[2])
+    vi_fr   = var(pm, nw, :vi, f_idx[2])
+    vr_to   = var(pm, nw, :vr, t_idx[2])
+    vi_to   = var(pm, nw, :vi, t_idx[2])
+
+    for (idx, f_terminal, t_terminal, r) in zip(1:length(f_connections), f_connections, t_connections, c_rating)
+        if r<Inf
+            JuMP.@constraint(pm.model, p_fr[idx]^2+q_fr[idx]^2 <= r^2*(vr_fr[f_terminal]^2+vi_fr[f_terminal]^2))
+            JuMP.@constraint(pm.model, p_to[idx]^2+q_to[idx]^2 <= r^2*(vr_to[t_terminal]^2+vi_to[t_terminal]^2))
+        end
```
pseudocubic, line 661: I suggest using `rating` instead of `r`, just for easier readability

src/form/en_acr.jl
```diff
@@ -0,0 +1,839 @@
...
+		report::Bool=true,
+		kwargs...
+	)
+
+For ACR models with explicit neutrals,
+creates branch power variables `:p` and `:q` and placeholder dictionaries for the terminal power flows `:p_bus` and `:q_bus`.
+"""
+function variable_mc_branch_power(pm::AbstractExplicitNeutralACRModel; nw::Int=nw_id_default, bounded::Bool=true, report::Bool=true, kwargs...)
+    variable_mc_branch_power_active(pm, nw=nw, bounded=bounded, report=report; kwargs...)
+    variable_mc_branch_power_reactive(pm, nw=nw, bounded=bounded, report=report; kwargs...)
```
pseudocubic, line 537: Should use `real` and `imaginary` instead of `active` and `reactive` for consistency with other naming conventions

src/form/en_acr.jl
```diff
@@ -0,0 +1,839 @@
...
+
+For ACR models with explicit neutrals,
+imposes a bound on the magnitude of the total apparent power at each winding of the transformer.
+
+```
+sum(pt_fr)^2 + sum(qt_fr)^2 <= sm_ub^2
+sum(pt_to)^2 + sum(qt_to)^2 <= sm_ub^2
+```
+"""
+function constraint_mc_transformer_power_rating(pm::AbstractExplicitNeutralACRModel, nw::Int, id::Int, f_idx::Tuple, t_idx::Tuple, f_bus::Int, t_bus::Int, f_connections::Vector, t_connections::Vector, config::ConnConfig, sm_ub::Real; report::Bool=true)
```
pseudocubic, line 506: Should be `thermal_rating` to be consistent with other naming conventions

pseudocubic:
Apologies for the delay in the review. I have quite a few comments, mostly either questions on functionality or comments on naming conventions. Overall this is looking in good shape, and I saw a decent number of bug fixes. Make sure that you are tracking those bug fixes for the CHANGELOG so that they get mentioned when we release the new version.

src/core/base.jl
```diff
@@ -86,23 +86,44 @@ end
...
-    set_lower_bound(x::JuMP.VariableRef, bound)
+	function set_lower_bound(
+		x::JuMP.VariableRef,
+		bound::Real
+	)
 
 Local wrapper method for JuMP.set_lower_bound, which skips NaN and infinite (-Inf only)
 """
-function set_lower_bound(x::JuMP.VariableRef, bound; loose_bounds::Bool=false, pm=missing, category::Symbol=:default)
+function set_lower_bound(x::JuMP.VariableRef, bound::Real)
```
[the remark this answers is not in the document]
sanderclaeys: I changed it also for `set_upper_bound`

src/data_model/eng2math.jl
```diff
@@ -457,6 +456,7 @@ function _map_eng2math_transformer!(data_math::Dict{String,<:Any}, data_eng::Dic
                     "polarity"      => eng_obj["polarity"][w],
                     "tm_set"        => eng_obj["tm_set"][w],
                     "tm_fix"        => eng_obj["tm_fix"][w],
+                    "sm_ub"         => get(eng_obj, "sm_ub", Inf),
```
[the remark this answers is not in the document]
sanderclaeys: I intended it to be `Real`. I discussed this a while back with @frederikgeth , and as a philosophy for bound specs we tried to stick as close as possible to how manufacturers supply ratings; for lines this means a per-conductor current rating, but for transformers it is a total apparent power rating. I kept on supporting the ones common in the OPF community as well (thermal rating for lines), but since these transformers bounds don't have strong convention, I went with the one that we believe makes most sense.

src/data_model/math2eng.jl
```diff
@@ -252,29 +252,22 @@ end
...
-
     if isa(map["to"], String)
         math_obj = _get_math_obj(data_math, map["to"])
         merge!(eng_obj, math_obj)
     else
         for to_id in map["to"]
             if startswith(to_id, "switch")
                 math_obj = _get_math_obj(data_math, to_id)
-                merge!(eng_obj, math_obj)
+                # skip to-side power and current; these come from branch
```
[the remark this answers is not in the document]
sanderclaeys: I reorganized this by importing everything from the switch first, and then importing the to-side power/current from the branch, which will overwrite the switch ones at that point. This is the desired behavior I believe.

src/core/objective.jl
```diff
@@ -273,6 +273,33 @@ function _objective_mc_min_fuel_cost_polynomial_linquad(pm::AbstractUnbalancedPo
...
+            end
+        end
+    end
+
+    return JuMP.@objective(pm.model, Min,
+        sum(
+            sum( gen_cost[(n,i)] for (i,gen) in nw_ref[:gen] )
+        for (n, nw_ref) in nws(pm))
+    )
+end
```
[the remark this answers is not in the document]
sanderclaeys: I merged all into a single function, which inspects whether `pg` contains any NL expressions and then acts accordingly. Typically we prefer doing this by overloading, but that does not work here because along the type hierarchy there is not a single branch which has NL expressions; i.e., the quadratic IVR does not have these, whilst the parent IVR has NL expressions, and this leads to a lot of code duplication. The change I made now prevents this.

src/core/objective.jl
```diff
@@ -331,6 +358,67 @@ function _objective_mc_min_fuel_cost_polynomial_linquad(pm::AbstractUnbalancedIV
...
+            end
+        end
+    end
+
+    return JuMP.@NLobjective(pm.model, Min,
+        sum(
+            sum(    gen_cost[(n,i)] for (i,gen) in nw_ref[:gen] )
+        for (n, nw_ref) in nws(pm))
+    )
+end
```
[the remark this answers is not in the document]
sanderclaeys: see reply there

src/core/objective.jl
```diff
@@ -331,6 +358,67 @@ function _objective_mc_min_fuel_cost_polynomial_linquad(pm::AbstractUnbalancedIV
...
```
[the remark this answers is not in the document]
sanderclaeys: see reply there

src/form/en_shared.jl
```diff
@@ -0,0 +1,1314 @@
...
+                                    + sum(crsw[a_sw][t] for (a_sw, conns) in bus_arcs_sw if t in conns)
+                                    + sum(crt[a_trans][t] for (a_trans, conns) in bus_arcs_trans if t in conns)
+                                    ==
+                                      sum(crg[g][t]         for (g, conns) in bus_gens if t in conns)
+                                    - sum(crs[s][t]         for (s, conns) in bus_storage if t in conns)
+                                    - sum(crd[d][t]         for (d, conns) in bus_loads if t in conns)
+                                    - sum( Gt[idx,jdx]*vr[u] -Bt[idx,jdx]*vi[u] for (jdx,u) in ungrounded_terminals) # shunts
+                                    )
+        # @smart_constraint(pm.model, [ci, cid, cig, cis, cisw, cit, vi],
+        @smart_constraint(pm.model, [ci, cid, cig, cit, vi],
```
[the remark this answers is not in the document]
sanderclaeys: good catch, this is a fragment left from debugging

src/form/en_ivr.jl
sanderclaeys: You mean `thermal_limit` right?

src/form/en_ivr.jl
sanderclaeys: This constraint is intended for branches, and not for a generic edge type. I did not include `branch` in the constraint name to be in line with previous naming conventions. Should I follow the existing naming scheme or correct it here to include the branch qualifier?

src/form/en_shared.jl
```diff
@@ -0,0 +1,1314 @@
...
+"""
+function constraint_mc_voltage_reference(pm::ExplicitNeutralModels, id::Int; nw::Int=nw_id_default, bounded::Bool=true, report::Bool=true, kwargs...)
+    bus = ref(pm, nw, :bus, id)
+    terminals = bus["terminals"]
+    grounded = bus["grounded"]
+
+    if haskey(bus, "va") && !haskey(bus, "vm")
+        constraint_mc_theta_ref(pm, id, nw=mw, bounded=bounded, report=report, kwargs...)
+    elseif haskey(bus, "vm") && !haskey(bus, "va")
+        error("Reference buses with magnitude-only (vm) setpoints are not supported. The same can be achieved by setting vmin=vmax=vm.")
```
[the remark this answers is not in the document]
sanderclaeys: You are right, I added a new function to do this

sanderclaeys:
@pseudocubic I resolved most comments, and left a few open where I asked for clarification or described how I tackled non-trivial issues, awaiting your further review.

Which bug fixes do you refer to, with respect to adding them to the changelog? Most of this PR is new functionality, so I can only think of some bug fixes in the previous PR which are still missing in the changelog. Do you refer to those maybe?

src/data_model/units.jl
```diff
@@ -362,6 +365,7 @@ end
 function _rebase_pu_branch!(branch::Dict{String,<:Any}, vbase::Real, sbase::Real, sbase_old::Real, voltage_scale_factor::Real)
     if !haskey(branch, "vbase")
         z_old = 1
+        vbase_old = 1.0
```
pseudocubic, line 368: Bug fix, include in changelog

src/data_model/units.jl
```diff
@@ -371,10 +375,12 @@ function _rebase_pu_branch!(branch::Dict{String,<:Any}, vbase::Real, sbase::Real
     z_scale = z_old/z_new
     y_scale = 1/z_scale
     sbase_scale = sbase_old/sbase
+    ibase_scale = sbase_scale/(vbase_old/vbase)
 
     _scale_props!(branch, ["br_r", "br_x"], z_scale)
     _scale_props!(branch, ["b_fr", "g_fr", "b_to", "g_to"], y_scale)
-    _scale_props!(branch, ["c_rating_a", "c_rating_b", "c_rating_c", "rate_a", "rate_b", "rate_c"], sbase_scale)
+    _scale_props!(branch, ["c_rating_a", "c_rating_b", "c_rating_c"], ibase_scale)
+    _scale_props!(branch, ["rate_a", "rate_b", "rate_c"], sbase_scale)
```
pseudocubic, line 383: Bug fix, include in changelog

src/data_model/units.jl
```diff
@@ -386,9 +392,13 @@ end
...
 "per-unit conversion for switches"
 function _rebase_pu_switch!(switch::Dict{String,<:Any}, vbase::Real, sbase::Real, sbase_old::Real, voltage_scale_factor::Real)
+    vbase_old = !haskey(switch, "vbase") ? 1.0 : switch["vbase"]
+    
     sbase_scale = sbase_old / sbase
+    ibase_scale = sbase_scale/(vbase_old/vbase)
 
-    _scale_props!(switch, ["c_rating_a", "c_rating_b", "c_rating_c", "rate_a", "rate_b", "rate_c"], sbase_scale)
+    _scale_props!(switch, ["current_rating", "c_rating_b", "c_rating_c"], ibase_scale)
+    _scale_props!(switch, ["thermal_rating", "rate_b", "rate_c"], sbase_scale)
```
pseudocubic, line 401: bug fix, include in changelog

src/data_model/utils.jl
```diff
@@ -400,6 +402,7 @@ function _build_loss_model!(
             "b_to" => diagm(0=>fill(b_to, nphases)),
             "angmin" => fill(-10.0, nphases),
             "angmax" => fill( 10.0, nphases),
+            "c_rating_a" => fill(Inf, nphases),
```
pseudocubic, line 405: We need to think about how to propagate bounds to transformer math objects, so that we can easily run bounds-requiring formulations without a weird workflow

src/data_model/utils.jl
```diff
@@ -957,14 +960,6 @@ function _add_implicit_absolute_bounds!(bus_math, terminals::Vector)
...
-        bus_math["vm_pair_lb"] = vm_pair_lb[lb_keep_idx]
-    end
-    if isempty(ub_keep_idx)
-        delete!(bus_math, "vm_pair_ub")
-    else
-        bus_math["vm_pair_ub"] = vm_pair_ub[ub_keep_idx]
-    end
+    bus_math["vm_pair_lb"] = vm_pair_lb[lb_keep_idx]
+    bus_math["vm_pair_ub"] = vm_pair_ub[ub_keep_idx]
 end
```
pseudocubic, line 965: Add line to EOF

src/form/utils.jl
```diff
@@ -0,0 +1,18 @@
...
+            flows_merged_t = flows_t[1]
+        elseif any(isa(a, JuMP.NonlinearExpression) for a in flows_t)
+            flows_merged_t = JuMP.@NLexpression(pm.model, sum(flows_t[i] for i in 1:length(flows_t)))
+        else
+            flows_merged_t = sum(flows_t)
+        end
+        push!(flows_merged, flows_merged_t)
+    end
+    JuMP.Containers.DenseAxisArray(flows_merged, conns_unique)
+end
```
pseudocubic, line 18: Add line to EOF

src/form/en_acr.jl
```diff
@@ -0,0 +1,840 @@
...
+"""
+function constraint_mc_generator_power_wye(pm::AbstractExplicitNeutralACRModel, nw::Int, id::Int, bus_id::Int, connections::Vector{Int}, pmin::Vector{<:Real}, pmax::Vector{<:Real}, qmin::Vector{<:Real}, qmax::Vector{<:Real}; report::Bool=true, bounded::Bool=true)
+    vr = var(pm, nw, :vr, bus_id)
+    vi = var(pm, nw, :vi, bus_id)
+    pg = var(pm, nw, :pg, id)
+    qg = var(pm, nw, :qg, id)
+
+    phases = connections[1:end-1]
+    n      = connections[end]
+    P      = length(phases)
```
pseudocubic, line 52: what happens if `connections` is not organized by `[phases..., neutral]`? E.g., in cases where we pad properties, i.e. on single phase delta components, the padded properties are added to the end, so `connections` could look like `[1,4,2,3]`. Is there a way we should keep track of which terminal is neutral other than assuming it is the last terminal?

pseudocubic (commented):
For the most part, still just minor comments, but there is one large issue that occurred to me as I was reviewing this. I only put the comment about it in one place, but it applies pretty globally to constraints, and that is, what if the `connections` vector is not correctly ordered such that the last conductor is not the neutral as is expected? In particular, when using the phase projections padding for single phase delta components, we can run into situations where the phase ordering becomes `[1,4,2,3]`. Is there a way to keep track of which conductors are neutrals and which are phases? Is there even a way to tell this by an opendss specification, or is any phase >3 assumed to be a neutral in opendss? 

Also, there is something failing in the Julia 1.0 tests

sanderclaeys:
I will write here my thoughts on the two main issues left in the PR to get your input before pushing an update.
## Infinite bounds
The convention we adopted a while back, is to set lower bounds to `-Inf` and upper bounds to `Inf` when they do not apply. For simple bounds this is resolved correctly by `set_lower_bound` and `set_upper_bound`, whilst for example for bounds of the form `x^2 + y^2 <= rating^2` this require an if-clause to skip the infinite ratings.

I added infinite bounds in several places, to make the data model transformation consistent with this convention. Already before, the data model transformation resulted in unbounded variables (now explicitly indicated with `Inf` bounds) which would lead to issues for certain formulations.

In my view, it would make sense to expect of the user to impose bounds manually at the `ENGINEERING` level. Automatically adding bounds generally is not a good practice. However, what is problematic is that even if all bounds are set at the `ENGINEERING` level, unbounded variables will appear because of the virtual buses, lines etc. Therefore, I propose that we automatically try to infer valid bounds for the virtual components, so that the user does not also have to add bounds manually for the virtual components. Such feature is beyond the scope of this PR, so I propose we add it after this PR. If you agree, I will create an issue for this and outline how to handle this.

## Neutral semantics
WIP

ccoffrin:
Some remarks on variable bounds in JuMP and solvers.  Solvers generally do not like `inf` values.  For solvers it is often better to specify no bounds, rather specify `inf` explicitly.  When I reasonable non-`inf` variable bound can be inferred, it is usually better for the solvers if this is added to the variable. Since I last checked JuMP may filter out `inf` bounds automatically when it sends the model to a solver, it is worth double checking that.

sanderclaeys:
@ccoffrin the `Inf` bounds never get passed to the solver, because
- we use a wrapper around `set_upper_bound` and `set_lower_bound` to ignore `Inf` bounds (I think this is obsolete now because this seems to have become default JuMP behavior, like you suggested);
- `Inf` values in constraint lead to an error, so even if an implementation forgets to handle `Inf` values it would not be passed to the solver silently.

With respect to adding valid bounds wherever possible, there is also a drawback. E.g., whilst `vm` bounds are simple bounds in polar coordinates, they lead to SoC constraints in rectangular coordinates. So automatically adding them wherever possible, can slow down some formulations. Personally I favor letting the modeler control which bounds are added (except perhaps the virtual components because of the discussion above).

pseudocubic:
@sanderclaeys I probably wasn't clear enough about the bounds issue. I wasn't really referring to default bounds coming from OpenDSS in this case, but only the second issue you noted, i.e., bounds on virtual objects, which have no easy control when working within the engineering model. With regards to the first part about default bounds, I still think we should add a helper function that users can use to set some reasonable default bounds, but I agree that because OpenDSS does not have clean ways to specify many bounds, we should not include them by default.

As to your suggestion: 

> Therefore, I propose that we automatically try to infer valid bounds for the virtual components, so that the user does not also have to add bounds manually for the virtual components. Such feature is beyond the scope of this PR, so I propose we add it after this PR. If you agree, I will create an issue for this and outline how to handle this.

This is acceptable to me, with the note that it is a high priority to fix this, mainly because certain relaxations require finite bounds to work at all, and therefore this needs to be addressed sooner rather than later.

With regards to the discussion on neutral semantics:

Padding is currently required for all _delta_ connected components in any formulation, and I have not been able to remove this requirement yet in the existing formulations. Did I miss that this has been addressed for the formulations you are adding, and if so would it be possible to extend this fix to the existing formulation in another PR? I note that padding is basically how Kersting describes how to model delta connected with <3 phases components. If there is a universal fix we can include for this, a helper function to note that 4 is not the last connection should suffice, because then the padding functions will not become the default, and will be only used by users who are interested.

sanderclaeys:
Agreed on the virtual component bounds; I will create an issue for it and it will be my next priority for PMD after this PR.

With respect to the padding, I think it would be helpful to understand what kind of delta-connected components we aim to support. Personally, I am concerned about the following two
- three-phase delta connections, i.e. for which all delta implementations are intended;
- two-phase delta connections, connected between two phase terminals.

Are there any other you would like to support? (There are open-delta connections etc., but these never were of interest to me)

pseudocubic:
Those two are the most common, and I know there is some interest in open-delta transformers from the protection team, though I don't know their priority for that

One note is that we should go back and make sure all "single-phase" delta connections are handled properly everywhere 

Correct me if I am mistaken, but my understanding is that a proper delta needs to be explicitly connected to at least 2 phases, but in the case where the second connection is a ground, when we kron reduce they get reduced to a single-phase connection, in which case we should model them as wye?

src/data_model/utils.jl
sanderclaeys: #351

src/data_model/eng2math.jl
```diff
@@ -798,6 +808,7 @@ function _map_eng2math_voltage_source!(data_math::Dict{String,<:Any}, data_eng::
                 "t_connections" => eng_obj["connections"],
                 "angmin" => fill(-10.0, nconductors),
                 "angmax" => fill( 10.0, nconductors),
+                "c_rating_a" => fill(Inf, nconductors),
```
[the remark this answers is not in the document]
sanderclaeys: #351

src/data_model/eng2math.jl
```diff
@@ -542,6 +543,7 @@ function _map_eng2math_switch!(data_math::Dict{String,<:Any}, data_eng::Dict{Str
                 "b_to" => zeros(nphases, nphases),
                 "angmin" => fill(-10.0, nphases),
                 "angmax" => fill( 10.0, nphases),
+                "c_rating_a" => fill(Inf, nphases),
```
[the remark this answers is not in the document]
sanderclaeys: #351

src/data_model/eng2math.jl
```diff
@@ -510,14 +508,17 @@ function _map_eng2math_switch!(data_math::Dict{String,<:Any}, data_eng::Dict{Str
...
                 "bus_i" => length(data_math["bus"])+1,
                 "bus_type" => get(eng_obj, "status", ENABLED) == DISABLED ? 4 : 1,
-                "terminals" => t_bus["terminals"],  # connected to the switch on the to-side
-                "grounded" => t_bus["grounded"],  # connected to the switch on the to-side
-                "vmin" => t_bus["vmin"],
-                "vmax" => t_bus["vmax"],
+                "terminals" => eng_obj["t_connections"],  # connected to the switch on the to-side
+                "grounded" => fill(false, N),  # connected to the switch on the to-side
+                "vmin" => fill(0.0, N),
+                "vmax" => fill(Inf, N),
```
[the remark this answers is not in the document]
sanderclaeys: #351

sanderclaeys:
I would not support the open-delta explicitly, as also OpenDSS does not do so either as far as I am aware. At most we would get an implicit open-delta as a composition of transformers, so we can handle that like that if needed.

Yes that's right. Actually, for the explicit neutral case, the distinction between phases and neutrals is not that meaningful, these are all simply conductors in the end. So a two-phase `DELTA` and a single-phase `WYE` are both two-conductor components. I support two-phase `DELTA` explicitly by using the relevant wye-connected component methods for them. Therefore, padding is not needed to support them implicitly.

This  is not possible for the Kron-reduced case like you noted. Under Kron-reduction, two-phase `DELTA` and single-phase `WYE` have different models. Since there is no explicit two-phase `DELTA` support, this is handled through padding.

More specifically to the issue you raised. Could we just skip padding any `WYE` component, or are there formulations that only support three-phase `WYE` components? Otherwise that is an easy way to avoid issues with inserting padded phases before the neutral terminal of `WYE` components.

An alternative would be to have the neutral as a separate property instead of as the last elements of conductors, but that would break a very generally applicable pattern and would be problematic for transformers where some windings have a neutral and others don't.

sanderclaeys:
(Since the four-wire formulations do not require padding, I would propose we move this PR along and further discuss the padding in a separate issue. I will solve the open minor comments tomorrow.)

src/form/en_acr.jl
sanderclaeys: See suggestions at the bottom; in my four-wire workflow I turn of the padding with the newly added flag for that.

src/data_model/units.jl
```diff
@@ -17,12 +17,13 @@ const dimensionalize_math = Dict{String,Dict{String,Vector{String}}}(
...
-        "ibase_to"=>Vector{String}(["crt_to", "cit_to"])
+        "sbase"=>Vector{String}(["pf", "qf", "pt", "qt"]),
+        "ibase_fr"=>Vector{String}(["cr_fr", "ci_fr"]),
+        "ibase_to"=>Vector{String}(["cr_to", "ci_to"])
     ),
     "switch" => Dict{String,Vector{String}}(
-        "sbase" => Vector{String}(["psw_fr", "psw_to", "qsw_fr", "qsw_to"]),
-        "ibase" => Vector{String}(["crsw_fr", "cisw_fr", "crsw_to", "cisw_to"])
+        "sbase" => Vector{String}(["pf", "pt", "qf", "qt"]),
+        "ibase" => Vector{String}(["cr_fr", "ci_fr", "cr_to", "ci_to"])
```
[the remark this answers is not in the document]
sanderclaeys: Do you mean in the changelog?

sanderclaeys:
Apparently the syntax `A.<:b` behaves weirdly in Julia `1.0`; it does not recognize b as a Type, but parses it as a Symbol.

 I rewrote it as a list comprehension `[a<:b for a in A]` so it now works for all versions.

sanderclaeys:
Now only one test in the latest Ubuntu version is failing... I will need to experiment through the CI commit check, since I do not have access to an Ubuntu environment

PR #342, opened 2021-07-07
ADD: Load models for LinDist3Flow

keatsig:
This adds the following load models to the LinDist3Flow formulation:

1. Wye-connected, constant current loads
2. Delta connected, constant power/current/impedance loads

### DIFF
--- a/src/form/bf_mx_lin.jl
+++ b/src/form/bf_mx_lin.jl
@@ -173,7 +173,21 @@ function constraint_mc_storage_losses(pm::AbstractUBFModels, i::Int; nw::Int=nw_
     )
 end
 
-## KIG modifications
+@doc raw"""
+    constraint_mc_load_power(pm::LPUBFDiagModel, load_id::Int; nw::Int=nw_id_default, report::Bool=true)
+
+Delta/voltage-dependent load models for LPUBFDiagModel. Delta loads use the auxilary power variable (X). The constant current load model is derived by linearizing around the flat-start voltage solution. 
+
+```math
+\begin{align}
+&\text{Constant power:} \Rightarrow P_i^d = P_i^{d0},~Q_i^d = Q_i^{d0} ~\forall i \in L \\
+&\text{Constant impedance (Wye):} \Rightarrow P_i^d = a_i \cdot w_i,~Q_i^d = b_i \cdot w_i ~\forall i \in L \\
+&\text{Constant impedance (Delta):} \Rightarrow P_i^d = 3\cdot a_i \cdot w_i,~Q_i^d = 3\cdot b_i \cdot w_i ~\forall i \in L \\
+&\text{Constant current (Wye):} \Rightarrow P_i^d = \frac{a_i}{2}\cdot \left( 1+w_i \right),~Q_i^d = \frac{b_i}{2}\cdot \left( 1+w_i \right) \forall i \in L \\
+&\text{Constant current (Delta):} \Rightarrow P_i^d = \frac{\sqrt{3} \cdot a_i}{2}\cdot \left( 1+w_i \right),~Q_i^d = \frac{\sqrt{3} \cdot b_i}{2}\cdot \left( 1+w_i \right) \forall i \in L 
+\end{align}
+```
+"""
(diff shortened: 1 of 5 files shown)


src/form/bf_mx_lin.jl
```diff
@@ -172,3 +172,101 @@ function constraint_mc_storage_losses(pm::AbstractUBFModels, i::Int; nw::Int=nw_
         qsc + q_loss
     )
 end
+
+## KIG modifications
```
pseudocubic, line 176: Should have a docstring of the format:

```julia
@doc raw"""
    constraint_mc_load_power(pm::LPUBFDiagModel, load_id::Int; nw::Int=nw_id_default, report::Bool=true)

Short description here

latex math equations here
"""
```

See https://github.com/lanl-ansi/PowerModelsONM.jl/blob/main/src/form/shared.jl#L1-L12 as an example

the `@doc raw"""` part is necessary for inserting math equations due to the usage of `\`, so that Julia doesn't throw an error about invalid escape sequences

src/form/bf_mx_lin.jl
```diff
@@ -172,3 +172,101 @@ function constraint_mc_storage_losses(pm::AbstractUBFModels, i::Int; nw::Int=nw_
...
+
+        ## reporting; for delta these are not available as saved variables!
+        if report
+            sol(pm, nw, :load, load_id)[:pd] = pd
+            sol(pm, nw, :load, load_id)[:qd] = qd
+            sol(pm, nw, :load, load_id)[:pd_bus] = pd_bus
+            sol(pm, nw, :load, load_id)[:qd_bus] = qd_bus
+        end
+    end
+end
```
pseudocubic, line 272: Files should end in an empty line

test/data/opendss/case3_unbalanced_delta_loads.dss
```diff
@@ -0,0 +1,40 @@
...
+New Load.L5 phases=1  loadbus.2.0   ( 0.4 3 sqrt / )   kW=6   kvar=3  model=2
+New Load.L6 phases=1  loadbus.3.0   ( 0.4 3 sqrt / )   kW=6   kvar=3  model=5
+
+
+Set voltagebases=[0.4]
+Set tolerance=0.000001
+set defaultbasefreq=50
+Calcvoltagebases
+
+Solve
```
pseudocubic, line 40: Files should end with an empty line

test/opf_bf.jl
```diff
@@ -49,6 +49,15 @@
...
             @test norm(sol["solution"]["bus"]["loadbus"]["w"]-[0.92234, 0.95778], Inf) <= 1E-4
         end
+        @testset "3-bus unbalanced lpubfdiag opf_bf with delta loads" begin
+            pmd = parse_file("../test/data/opendss/case3_unbalanced_delta_loads.dss")
+            apply_voltage_bounds!(pmd; vm_lb=0.9, vm_ub=1.1);
+            sol = solve_mc_opf(pmd, LPUBFDiagPowerModel, ipopt_solver; make_si=false)
+            @test sol["termination_status"] == LOCALLY_SOLVED
+            baseMVA = sol["solution"]["settings"]["sbase"] / sol["solution"]["settings"]["power_scale_factor"]
+            @test isapprox(sum(sol["solution"]["voltage_source"]["source"]["pg"] * baseMVA), 0.04026874; atol=2e-3)
+            @test isapprox(sum(sol["solution"]["voltage_source"]["source"]["qg"] * baseMVA), 0.0171721; atol=2e-3)
```
pseudocubic, line 59: Q: are the power values being compared here from OpenDSS?

I think also we should add some tests of the voltages at at least one of the buses, probably at `loadbus`.

pseudocubic (changes requested):
Overall the implementation looks good to me, and the code is in really good shape. 

I left some comments, but to summarize the changes I think are still needed:

- Documentation for the new constraint should be added, 
- I think we should add an additional check to the unit tests, and 
- CHANGELOG.md should be updated (add a bullet under `## staged` with a short description of the update)

test/opf_bf.jl
keatsig: Yes, the power values are compared with the OpenDSS power flow solution.

pseudocubic (approved):

PR #347, opened 2021-07-19
ADD: Forward-backward sweep (FBS) linear formulation

keatsig:
This adds a forward-backward sweep (FBS) linear unbalanced branch flow formulation. The model is linearized around an initial operating point using a single iteration of the FBS method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## staged
 
+- Added simplified linear UBF formulation using a single iteration of forward-backward sweep (FBS)
 - Added support for raw parsing of RegControl and CapControl objects from dss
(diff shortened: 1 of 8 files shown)


pseudocubic (commented):
It looks like all my previous offline comments were covered. I thought of two more minor things we should do:

- Add a brief bit on FBS formulation to the [formations documentation](https://github.com/lanl-ansi/PowerModelsDistribution.jl/blob/main/docs/src/manual/formulations.md)
- Change the filename from `bf_mx_fbs.jl` to `bf_fbs.jl`; it doesn't really use the same function patterns as the other functions in `bf_mx.jl`

pseudocubic (approved):

PR #349, opened 2021-07-28
ADD: Voltage regulator control 

keatsig:
This adds voltage regulator control to the following formulations:

1. Nonlinear: ACP and ACR
2. Linear: LinDist3Flow and FBS.

This control emulates a typical utility wye-connected voltage regulator by changing the transformer taps based on the RegControl dss objects.


src/data_model/eng2math.jl
```diff
@@ -412,7 +412,7 @@ function _map_eng2math_transformer!(data_math::Dict{String,<:Any}, data_eng::Dic
         else
             vnom = eng_obj["vm_nom"] * data_eng["settings"]["voltage_scale_factor"]
             snom = eng_obj["sm_nom"] * data_eng["settings"]["power_scale_factor"]
-
+
```
pseudocubic, line 415: Extraneous spaces

src/io/dss/dss2eng.jl
```diff
@@ -725,6 +734,21 @@ function _dss2eng_transformer!(data_eng::Dict{String,<:Any}, data_dss::Dict{Stri
...
+            defaults = _apply_ordered_properties(_create_regcontrol(id; _to_kwargs(reg_ctrl)...), reg_ctrl)
+            
+            eng_obj["controls"] = Dict{String,Any}(
+                "name" => reg_ctrl["name"],
+                "vreg" => defaults["vreg"],
+                "band" => defaults["band"],
+                "ptratio" => defaults["ptratio"],
+                "ctprim" => defaults["ctprim"],
+                "r" => defaults["r"],
+                "x" => defaults["x"]
```
pseudocubic, line 749: these need to be updated to vectors as in the documentation, using the `dss_obj["winding"]` to determine which winding is being controlled. Also, probably `tm_fixed` on the transformer should be updated according to which winding is being controlled?

src/io/dss/dss2eng.jl
```diff
@@ -547,6 +547,15 @@ end
...
 function _dss2eng_transformer!(data_eng::Dict{String,<:Any}, data_dss::Dict{String,<:Any}, import_all::Bool)
+    # check if regcontrol items are present
+    reg_obj = Dict()
+    if haskey(data_dss, "regcontrol")
+        for (id, dss_obj) in get(data_dss, "regcontrol", Dict{String,Any}())
+            reg_obj["$(dss_obj["transformer"])"] = dss_obj
+            reg_obj["$(dss_obj["transformer"])"]["name"] = id
+        end
+    end
+
```
pseudocubic, line 558: you should still do this in a separate `_dss2eng_regcontrol!`, so that you don't have to iterate over dict values so many times

```julia
function _dss2eng_regcontrol!(data_eng::Dict{String,<:Any}, data_dss::Dict{String,<:Any}, import_all::Bool)
    for (id, dss_obj) in get(data_dss), "regcontrol", Dict{String,Any}())
        # build "controls" dict eng_obj here
        data_eng["transformer"]["$(dss_obj["transformer"])"]["controls"] = eng_obj
    end
end
```

src/io/dss/dss2eng.jl
```diff
@@ -725,6 +734,21 @@ function _dss2eng_transformer!(data_eng::Dict{String,<:Any}, data_dss::Dict{Stri
...
+            eng_obj["controls"] = Dict{String,Any}(
+                "name" => reg_ctrl["name"],
+                "vreg" => defaults["vreg"],
+                "band" => defaults["band"],
+                "ptratio" => defaults["ptratio"],
+                "ctprim" => defaults["ctprim"],
+                "r" => defaults["r"],
+                "x" => defaults["x"]
+            ) 
+        end
```
pseudocubic, line 751: Should be moved into the `_dss2eng_regcontrol!` function.

src/io/utils.jl
```diff
@@ -370,6 +370,11 @@ function _bank_transformers!(data_eng::Dict{String,<:Any})
             btrans["status"] = all(tr["status"] == ENABLED for tr in trs)
             btrans["source_id"] = "transformer.$bank"
 
+            # add regulator objects if present
+            if haskey(trs[1],"controls")
+                btrans["controls"] = trs[1]["controls"]
+            end
```
pseudocubic, line 376: If the transformer is a bank we need to be careful of cases where only one phase of the bank is being controlled, and therefore vreg, band, ptratio, ctprim , r, and x, are different for each phase. This might mean that the controls need to be `Vector{Vector{Real}}`, like many of the transformer properties?

src/data_model/eng2math.jl
```diff
@@ -469,6 +469,11 @@ function _map_eng2math_transformer!(data_math::Dict{String,<:Any}, data_eng::Dic
 
                 data_math["transformer"]["$(transformer_2wa_obj["index"])"] = transformer_2wa_obj
 
+                # add regcontrol items to math model
+                if haskey(eng_obj,"controls") && eng_obj["bus"][2]==eng_obj["bus"][w]
+                    data_math["transformer"]["$(transformer_2wa_obj["index"])"]["controls"] = eng_obj["controls"]
+                    data_math["transformer"]["$(transformer_2wa_obj["index"])"]["tm_fix"] = fill(false,length(transformer_2wa_obj["f_connections"]))
+                end
```
pseudocubic, line 476: does this always mean the second winding is the one being controlled, and all phases are controlled, even in cases where there might not be controls for each phase (e.g., in cases where transformers are banks, and a regcontrol doesn't exist for each object in the bank)?

pseudocubic (approved):

PR #350, opened 2021-08-04
ADD: First-order Taylor (FOT) linear formulation

keatsig:
This adds a first-order Taylor (FOT) linear unbalanced formulation. The model is linearized around an initial operating point using the FOT approximation.


pseudocubic (approved):
No comments, looks in good shape

PR #356, opened 2021-08-27
[ADD] Explicit neutral models

sanderclaeys:
In order to bring this up to speed with `main`, I had to fix one test. A test had been added for switches in the meanwhile. This was breaking because now a `thermal_rating` is always present, possibly initialized to `Inf`, and `constraint_mc_switch_thermal_limit` was missing for `AbstractLPUBFModel`. I added a stub which throws a warning if finite limits are present.

For the unit testing, I included two types of tests
- `en_pf_validation.jl`: this validates the power flow results against OpenDSS results. I added two new folders to `data`, which contain `dss` cases and their solutions. Each case is a small as possible, and contains a specific modelling feature (e.g., a dy transformer). I included a script in the solutions folder, `generate_solutions.jl`. This uses OpenDSSDirect.jl to generate the solution files. By storing the solutions, we can prevent having to use OpenDSSDirect.jl in the unit tests directly.
- `en_opf_bounds.jl`: this test whether various bounds are correctly represented in each formulation.


pseudocubic:
I still have a little bit left to review, but I wanted to leave one comment in the meantime. This is looking really good to me so far, but we need a bit of documentation to be added to docs/src on the models, and it would be nice to have an additional pluto notebook for examples/ that demonstrates the usage of the new models

sanderclaeys:
Sounds good! I will make such a notebook and add the docs. 

Once I know how long I will need for that (given my other tasks), I will give you an update when you can expect these to be committed.

sanderclaeys:
I am currently on holidays until the 19th of September. I will carry out any changes / additions you requested in the week after that.

sanderclaeys:
I nearly finished the notebook. Should I enter `0.12` as a PMD version in there, i.e. for the first version which will include the EN models?

sanderclaeys:
@pseudocubic awaiting further input

pseudocubic:
I have reviewed again, and I think this looks done now; there are a couple of minor merge conflicts, but I will take care of them.

Before merging into main, I think I need one more minor 0.11 release to include a bugfix I'm working on right now, but I think you can expect me to merge this by the end of the week.

Thanks so much for your hard work @sanderclaeys, this is an excellent addition to PMD!

sanderclaeys:
Thank you @pseudocubic , always a pleasure to work with you :)

PR #402, opened 2022-10-11
FIX: indexing of single-phase delta loads for linear formulations

keatsig:
This PR fixes indexing issue for single-phase delta load models used in linear formulations (LinDist3Flow, FOTP, FOTR, FBS).

### DIFF
--- a/src/data_model/utils.jl
+++ b/src/data_model/utils.jl
@@ -535,7 +535,7 @@ function _pad_properties_delta!(object::Dict{String,<:Any}, properties::Vector{S
             merge!(tmp, Dict((k[2], k[1])=>sign*v for (k,v) in tmp))
             get_val(x,y) = haskey(tmp, (x,y)) ? tmp[(x,y)] : 0.0
 
-            object[property] = [get_val(phases[1], phases[2]), get_val(phases[2], phases[3]), get_val(phases[3], phases[1])]
+            object[property] = val_length==1 ? [get_val(connections[1], connections[2]), 0.0,0.0] : [get_val(phases[1], phases[2]), get_val(phases[2], phases[3]), get_val(phases[3], phases[1])]
         end
     end
 end
--- a/src/form/bf_mx_lin.jl
+++ b/src/form/bf_mx_lin.jl
@@ -438,13 +438,13 @@ function constraint_mc_load_power(pm::LPUBFDiagModel, load_id::Int; nw::Int=nw_i
                 JuMP.@constraint(pm.model, qd[idx]==qd0[idx])
             end
         elseif load["model"]==IMPEDANCE
-            w = var(pm, nw, :w)[bus_id][[c for c in connections]]
+            w = var(pm, nw, :w)[bus_id]
             for (idx,c) in enumerate(connections)
                 JuMP.@constraint(pm.model, pd[idx]==3*a[idx]*w[c])
                 JuMP.@constraint(pm.model, qd[idx]==3*b[idx]*w[c])
             end
         else
-            w = var(pm, nw, :w)[bus_id][[c for c in connections]]
+            w = var(pm, nw, :w)[bus_id]
             for (idx,c) in enumerate(connections)
                 JuMP.@constraint(pm.model, pd[idx]==sqrt(3)/2*a[idx]*(w[c]+1))
                 JuMP.@constraint(pm.model, qd[idx]==sqrt(3)/2*b[idx]*(w[c]+1))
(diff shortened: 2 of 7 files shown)


src/form/bf_mx_lin.jl
```diff
@@ -439,21 +440,21 @@ function constraint_mc_load_power(pm::LPUBFDiagModel, load_id::Int; nw::Int=nw_i
         elseif load["model"]==IMPEDANCE
             w = var(pm, nw, :w)[bus_id][[c for c in connections]]
             for (idx,c) in enumerate(connections)
-                JuMP.@constraint(pm.model, pd[idx]==3*a[idx]*w[idx])
-                JuMP.@constraint(pm.model, qd[idx]==3*b[idx]*w[idx])
+                JuMP.@constraint(pm.model, pd[idx]==3*a[idx]*w[c])
+                JuMP.@constraint(pm.model, qd[idx]==3*b[idx]*w[c])
```
pseudocubic, line 444: I'm not sure this is correct, because of how `w` is created in line 441; it should be a vector in the order of connections, and therefore should still be indexed by `idx`

src/form/bf_mx_lin.jl
```diff
@@ -439,21 +440,21 @@ function constraint_mc_load_power(pm::LPUBFDiagModel, load_id::Int; nw::Int=nw_i
...
             end
         else
             w = var(pm, nw, :w)[bus_id][[c for c in connections]]
             for (idx,c) in enumerate(connections)
-                JuMP.@constraint(pm.model, pd[idx]==sqrt(3)/2*a[idx]*(w[idx]+1))
-                JuMP.@constraint(pm.model, qd[idx]==sqrt(3)/2*b[idx]*(w[idx]+1))
+                JuMP.@constraint(pm.model, pd[idx]==sqrt(3)/2*a[idx]*(w[c]+1))
+                JuMP.@constraint(pm.model, qd[idx]==sqrt(3)/2*b[idx]*(w[c]+1))
```
pseudocubic, line 450: I'm not sure this is correct, because of how `w` is created in line 447; it should be a vector in the order of connections, and therefore should still be indexed by `idx`

src/data_model/utils.jl
```diff
@@ -535,7 +535,7 @@ function _pad_properties_delta!(object::Dict{String,<:Any}, properties::Vector{S
             merge!(tmp, Dict((k[2], k[1])=>sign*v for (k,v) in tmp))
             get_val(x,y) = haskey(tmp, (x,y)) ? tmp[(x,y)] : 0.0
 
-            object[property] = [get_val(phases[1], phases[2]), get_val(phases[2], phases[3]), get_val(phases[3], phases[1])]
+            object[property] = val_length==1 ? [get_val(connections[1], connections[2]), 0.0,0.0] : [get_val(phases[1], phases[2]), get_val(phases[2], phases[3]), get_val(phases[3], phases[1])]
```
pseudocubic, line 538: I'm not clear on why there are four numbers for the `val_length == 1` case, instead of 3 like in the original case

pseudocubic (changes requested):

src/data_model/utils.jl
keatsig: `get_val(x,y)` gives one number, which is basically `tmp[(connections[1], connections[2])]` (or `val[1]` as per line 528). The next two numbers are always zeros for single-phase delta loads. So, I think `object[property]` should still have only three numbers.

src/form/bf_mx_lin.jl
keatsig: I did a quick check by adding `@show pd[idx],qd[idx],a[idx],b[idx],w[c]` inside the for loop and the expressions used for each term are consistent with each other for all three iterations of the loop. If `w[idx]` is used, the order does not seem right. 

Actually, replacing `w = var(pm, nw, :w)[bus_id][[c for c in connections]]` with `w = var(pm, nw, :w)[bus_id]` without changing anything else also keeps the expressions consistent inside the for loop and gives the exact same results!

src/form/bf_mx_lin.jl
keatsig: Same issue as above. If `w[idx]` is used, the expressions are not consistent inside the for loop.

src/form/bf_mx_lin.jl
pseudocubic: Lets make that change to `w` then, i think it would be better, and should be guaranteed to be consistent in edge cases where the order is unusual.

src/form/bf_mx_lin.jl
pseudocubic: Same as above, lets change the definition of `w`

pseudocubic (approved):

PR #404, opened 2022-10-25
ADD: Capacitor control for IVR formulation

keatsig:
This PR adds capacitor control to IVR formulation. This control emulates a typical utility wye-connected capacitor by sending switching messages.

### DIFF
--- a/src/form/ivr.jl
+++ b/src/form/ivr.jl
@@ -191,6 +191,16 @@ function variable_mc_bus_voltage(pm::AbstractUnbalancedIVRModel; nw::Int=nw_id_d
 end
 
 
+"""
+    variable_mc_capcontrol(pm::AbstractUnbalancedIVRModel; nw::Int=nw_id_default, relax::Bool=false, report::Bool=true)
+
+Capacitor switching variables.
+"""
+function variable_mc_capcontrol(pm::AbstractUnbalancedIVRModel; nw::Int=nw_id_default, relax::Bool=false, report::Bool=true)
+    variable_mc_capacitor_switch_state(pm; nw=nw, relax=relax, report=report)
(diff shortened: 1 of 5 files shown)


src/form/ivr.jl
```diff
@@ -191,6 +191,16 @@ function variable_mc_bus_voltage(pm::AbstractUnbalancedIVRModel; nw::Int=nw_id_d
...
+"""
+    variable_mc_capcontrol(pm::AbstractUnbalancedIVRModel; nw::Int=nw_id_default, relax::Bool=false, report::Bool=true)
+
+Capacitor switching variables.
+"""
+function variable_mc_capcontrol(pm::AbstractUnbalancedIVRModel; nw::Int=nw_id_default, relax::Bool=false, report::Bool=true)
+    variable_mc_capacitor_switch_state(pm; nw=nw, relax=relax, report=report)
+end
+
+
```
pseudocubic, line 203: Is this necessary? Is having a function with `pm::AbstractUnbalancedPowerModels` would be enough, I don't see any custom version of `variable_mc_capacitor_switch_state` for IVR

keatsig: It might not be necessary. I think the other formulations (ACP, ACR) also have individual `variable_mc_capcontrol` but call the same `variable_mc_capacitor_switch_state` function. I can modify this and create a single `variable_mc_capcontrol(pm::AbstractUnbalancedPowerModels;...)` that should work for all formulations.

pseudocubic: yes, let's do that. the current way seems redundant, and this would reduce the amount of code we need to maintain in the future

pseudocubic (approved):

PR #422, opened 2023-01-12
FIX voltage source impedance for 4 wire networks

hei06j:
This PR modifies the voltage source impedance matrix to support 4-wire networks similar to OpenDSS. The modifications do not impact the 3-wire networks.

Key changes:
 - zm (mutual impedance) and zs (self impedance) are stored.
 - voltage source impedance matrix is updated.
 - one OpenDSS file in the tests is updated.
 
This PR will help with the PR on the native power flow solver.


pseudocubic:
Fixed by https://github.com/lanl-ansi/PowerModelsDistribution.jl/commit/2ca704d2270bd386f1c6c9351838b6bde8e3663d

PR #431, opened 2023-04-26
FIX: implementation of polynomial nl costs above quadratic

juanjospina:
This PR fixes the calculation of polynomial nonlinear costs above quadratic. Non-breaking fix.


juanjospina:
Thanks @ccoffrin. I saw your fix on PowerModels (double check it just to be sure), and then ported the fix to both PowerModelsDistribution and PowerModelsITD.

ccoffrin:
The fix looks correct.  I would suggest adding a new test so that there is
no regression during refactoring.  PowerModels has one such example

pseudocubic (approved):

PR #454, opened 2024-01-30
UPD: update to the NonlinearExpr syntax of JuMP v1.15

### LINKED ISSUE #453: Update to JuMP NonlinearExpr syntax
x-ref https://github.com/lanl-ansi/PowerModels.jl/pull/901

I'm working on this now.

odow:
Closes #453

The tests haven't finished running locally, so there might be some more changes needed, but this is a start. I'll comment in-line about some formatting changes we could do here, or in a separate PR.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -23,7 +23,6 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 CSV = "0.8.5, 0.9, 0.10"
-Dates = "1.6"
 FilePaths = "0.8.3"
 Glob = "1.3"
 Graphs = "1"
@@ -34,9 +33,7 @@ JuMP = "1.23.2"
 LoggingExtras = "0.4.7, 1"
 PolyhedralRelaxations = "0.3.5"
 SCS = "0.9, 1.0, 1.1"
-SparseArrays = "1.6"
 SpecialFunctions = "2"
-Test = "1.6"
 julia = "1.6"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -2,7 +2,7 @@ name = "PowerModelsDistribution"
 uuid = "d7431456-977f-11e9-2de3-97ff7677985e"
 authors = ["David M Fobes <dfobes@lanl.gov>", "Carleton Coffrin"]
 repo = "https://github.com/lanl-ansi/PowerModelsDistribution.jl.git"
-version = "0.15.2"
+version = "0.16.0"
 
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
@@ -23,22 +23,24 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 CSV = "0.8.5, 0.9, 0.10"
+Dates = "1.6"
 FilePaths = "0.8.3"
 Glob = "1.3"
 Graphs = "1"
 InfrastructureModels = "0.7.3, 0.7.5"
-Ipopt = "0.9, 1.0.2, 1.1"
+Ipopt = "1"
 JSON = "0.18, 0.19, 0.20, 0.21"
-JuMP = "0.22, 0.23, 1"
+JuMP = "1.23.2"
 LoggingExtras = "0.4.7, 1"
 PolyhedralRelaxations = "0.3.5"
 SCS = "0.9, 1.0, 1.1"
+SparseArrays = "1.6"
 SpecialFunctions = "2"
+Test = "1.6"
 julia = "1.6"
 
 [extras]
 Ipopt = "b6b21f68-93f8-5de0-b562-5493be1d77c9"
-JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 SCS = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
--- a/src/core/base.jl
+++ b/src/core/base.jl
@@ -70,22 +70,6 @@ end
 
 
 """
-    @smart_constraint model::JuMP.Model vars::Vector expr::JuMP.Expression
-
-Detection of whether a constraint should be NL or not"
-"""
-macro smart_constraint(model, vars, expr)
-    esc(quote
-        if _has_nl_expression($vars)
-            JuMP.@NLconstraint($model, $expr)
-        else
-            JuMP.@constraint($model, $expr)
-        end
-    end)
-end
-
-
-"""
 	function set_lower_bound(
 		x::JuMP.VariableRef,
 		bound::Real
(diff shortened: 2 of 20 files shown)


src/core/base.jl
```diff
@@ -69,22 +69,6 @@ function _check_var_keys(vars, keys, var_name, comp_name)
...
-"""
-macro smart_constraint(model, vars, expr)
-    esc(quote
-        if _has_nl_expression($vars)
-            JuMP.@NLconstraint($model, $expr)
-        else
-            JuMP.@constraint($model, $expr)
-        end
-    end)
-end
```
odow, line 85: This PR is breaking because it removes this macro which was documented.

src/form/en_shared.jl
```diff
@@ -133,7 +133,7 @@ function constraint_mc_current_balance(pm::RectangularVoltageExplicitNeutralMode
     ungrounded_terminals = [(idx,t) for (idx,t) in enumerate(terminals) if !grounded[idx]]
 
     for (idx, t) in ungrounded_terminals
-        @smart_constraint(pm.model,  [cr, crd, crg, crs, crsw, crt, vr],
+        JuMP.@constraint(pm.model,
```
odow, line 136: As currently formatted, the body of the constraint is indented by a lot. (It used to match the opening `[` of the second argument. I can either leave to minimize the diff of this PR, or fix. Not sure which is best.

odow:
@ccoffrin (or anyone): are the failing tests global solutions (and there's a bug somewhere)? Or are these local minima, and by changing the order of constraints etc we now find new solutions?

ccoffrin:
Good question. I am not that familiar with how stable the tests are in terms of local minima. The failing tests do appear to be consolidated in the non-convex formulations, so it is a reasonable hypothesis.  We will need one of the primary developers to look into it to assess.

odow:
Hey @pseudocubic any chance you could take a look at this? Or point me in the right direction?

pseudocubic:
@odow we have been looking at this but still haven't identified an obvious pattern in the failing tests, and based on the changes in PMD and JuMP 1.15 don't understand what would have caused all the failures we're seeing.

Project.toml
```diff
@@ -23,22 +23,24 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
...
+Dates = "1.6"
 FilePaths = "0.8.3"
 Glob = "1.3"
 Graphs = "1"
 InfrastructureModels = "0.7.3, 0.7.5"
-Ipopt = "0.9, 1.0.2, 1.1"
+Ipopt = "1"
 JSON = "0.18, 0.19, 0.20, 0.21"
-JuMP = "0.22, 0.23, 1"
+JuMP = "1.15"
```
odow, line 33:
```suggestion
JuMP = "1.23.2"
```

odow: Waiting for https://github.com/JuliaRegistries/General/pull/115113

odow:
Now we're getting somewhere. These two failures just look flakey? One looks like a tolerance issue, and the other is Ipopt not succeeding.

test/transformer.jl
```diff
@@ -369,7 +369,7 @@
 
         @testset "3w_dyy_1" begin
             result = solve_mc_opf(ut_trans_3w_dyy_1, SOCConicUBFPowerModel, scs_solver; solution_processors=[sol_data_model!], make_si=false)
-            @test norm(result["solution"]["bus"]["3"]["vm"]-[0.93180, 0.88827, 0.88581], Inf) <= 7.2E-2
+            @test norm(result["solution"]["bus"]["3"]["vm"]-[0.93180, 0.88827, 0.88581], Inf) <= 1E-3
```
odow, line 372:
```suggestion
            @test norm(result["solution"]["bus"]["3"]["vm"]-[0.93180, 0.88827, 0.88581], Inf) <= 1E-1
```

test/transformer.jl
```diff
@@ -369,7 +369,7 @@
 
         @testset "3w_dyy_1" begin
             result = solve_mc_opf(ut_trans_3w_dyy_1, SOCConicUBFPowerModel, scs_solver; solution_processors=[sol_data_model!], make_si=false)
-            @test norm(result["solution"]["bus"]["3"]["vm"]-[0.93180, 0.88827, 0.88581], Inf) <= 7.2E-2
+            @test norm(result["solution"]["bus"]["3"]["vm"]-[0.93180, 0.88827, 0.88581], Inf) <= 1E-1
```
odow, line 372: @pseudocubic I relaxed this tolerance.  The other tests above this also use a mix of tolerances, so it seems a bit flakey.

test/en_opf_bounds.jl
```diff
@@ -76,7 +76,9 @@ end
         # IVRReducedQuadraticENPowerModel does not implement sm_ub
 
         # ACRENPowerModel
-        sol_pmd = calc_sol_pmd(data_math, ACRENPowerModel; optimizer=optimizer_with_attributes(Ipopt.Optimizer, "mu_strategy"=>"adaptive", "sb"=>"yes", "print_level"=>0))
+        # This test is flakey.
+        mu_strategy = Sys.islinux() ? "adaptive" : "monotone"
+        sol_pmd = calc_sol_pmd(data_math, ACRENPowerModel; optimizer=optimizer_with_attributes(Ipopt.Optimizer, "mu_strategy"=>mu_strategy, "sb"=>"yes", "print_level"=>0))
```
odow, line 81: @pseudocubic I changed the `mu_strategy` to the default except for linux because this test seemed to fail sporadically. I assume that the non-default option was added precisely because it's required sometimes (but not all the time).

ccoffrin:
Given all tests are passing, is there any reason not to merge and tag?

odow:
From my side, I'm good.

Project.toml
```diff
@@ -23,22 +23,24 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
...
+Ipopt = "1"
 JSON = "0.18, 0.19, 0.20, 0.21"
-JuMP = "0.22, 0.23, 1"
+JuMP = "1.23.2"
 LoggingExtras = "0.4.7, 1"
 PolyhedralRelaxations = "0.3.5"
 SCS = "0.9, 1.0, 1.1"
+SparseArrays = "1.6"
 SpecialFunctions = "2"
+Test = "1.6"
```
pseudocubic, line 39: Is there a reason Dates, SparseArrays and Test were added to compat? My impression was that these don't need to be explicitly listed because of the julia compat entry.

pseudocubic (changes requested):
I have re-reviewed the PR from scratch, and left a comment on some lines added to Project.toml. 

Also, please add an update to the CHANGELOG

Otherwise, it looks good, thanks for helping us to track down the issue in JuMP and make these changes to PMD!

odow:
I've added the CHANGELOG entry.

For the compat bounds: future versions of Julia are going to decouple stdlib versions from the Julia version, so best practice is to add compat bounds. But there was a bug in Julia 1.6 that prevented them from enforcing this rule all packages at registration time.

I've removed for now, but in some future version, we'll likely need to add them back. (Probably when/if you set `julia = "1.10"`.)

pseudocubic (approved):

pseudocubic:
Looks good! Thanks for the info about stdlib versions getting decoupled, I didn't know that, so I'll keep an eye out in the future.

Thanks again for the contribution!

PR #457, opened 2024-02-16
[add] solve_mn_mc_opf_oltc

John-Boik:
# Pull Request (PR) Template

Pull requests should be single-issue to the extent possible; by focusing a PR on a single issue/topic, code reviews are more manageable, new features are easier to track, code quality is easier to maintain, and technical debt is easier manage.

Every PR to PowerModelsDistribution should strive to meet the following guidelines.

## Title
[ADD] function `solve_mn_mc_opf_oltc`

## Body
Non-breaking. The purpose is to extend `solve_mc_opf_oltc` to the multi-network setting. 

## Code
There are three changes:

- Add function `solve_mn_mc_opf_oltc` to src/prob/opf_oltc.jl
- Add function `build_mn_mc_opf_oltc` to src/prob/opf_oltc.jl. I modeled it on `build_mn_mc_opf`.
- Alter function `variable_mc_oltc_transformer_tap` in src/core/variable.jl to allow for nonexistent transformer keys `tm_lb` and `tm-ub`. I tested `solve_mn_mc_opf_oltc` using IEEE123 parsed from OpenDSS format. The parse did not set values for `tm_lb` and `tm-ub` for the transformers, and without this change the code throws missing key errors. This might not be the best way to handle the missing key issue, as now the default keyword `bounded=true` in the call to `variable_mc_oltc_transformer_tap` might have less meaning. 
- CHANGELOG.md was altered 
 
No documentation was changed, as the sister function `solve_mn_mc_opf` has no documentation. No unit tests were added, as there are no unit tests for the sister function `solve_mc_opf_oltc`.

### DIFF
--- a/src/core/variable.jl
+++ b/src/core/variable.jl
@@ -507,8 +507,10 @@ function variable_mc_oltc_transformer_tap(pm::AbstractUnbalancedPowerModel; nw::
 
     if bounded
         for tr_id in p_oltc_ids, p in 1:length(ref(pm,nw,:transformer,tr_id,"f_connections"))
-            set_lower_bound(var(pm, nw)[:tap][tr_id][p], ref(pm, nw, :transformer, tr_id, "tm_lb")[p])
-            set_upper_bound(var(pm, nw)[:tap][tr_id][p], ref(pm, nw, :transformer, tr_id, "tm_ub")[p])
+            if haskey(ref(pm, nw, :transformer, tr_id), "tm_lb") && haskey(ref(pm, nw, :transformer, tr_id), "tm_ub")
+                set_lower_bound(var(pm, nw)[:tap][tr_id][p], ref(pm, nw, :transformer, tr_id, "tm_lb")[p])
+                set_upper_bound(var(pm, nw)[:tap][tr_id][p], ref(pm, nw, :transformer, tr_id, "tm_ub")[p])
+            end
         end
     end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## staged
 
+- Added function `solve_mn_mc_opf_oltc` for multi-networks
 - Fixed bug in `DssLine` parser where `c1` was being set to `c0`
 - Fixed native pf unit tests, which assume no virtual branches from switches (applied `make_lossless!` before test)
 - Added `g_fr`, `g_to`, `b_fr`, `b_to` to switches in `dss2eng` and `eng2math`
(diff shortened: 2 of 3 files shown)


src/core/variable.jl
```diff
@@ -507,8 +507,10 @@ function variable_mc_oltc_transformer_tap(pm::AbstractUnbalancedPowerModel; nw::
 
     if bounded
         for tr_id in p_oltc_ids, p in 1:length(ref(pm,nw,:transformer,tr_id,"f_connections"))
-            set_lower_bound(var(pm, nw)[:tap][tr_id][p], ref(pm, nw, :transformer, tr_id, "tm_lb")[p])
-            set_upper_bound(var(pm, nw)[:tap][tr_id][p], ref(pm, nw, :transformer, tr_id, "tm_ub")[p])
+            if haskey(ref(pm, nw, :transformer, tr_id), "tm_lb") && haskey(ref(pm, nw, :transformer, tr_id), "tm_ub")
```
pseudocubic, line 510: I am not sure why this check is necessary, because in principle tm_lb and tm_ub should always exist in the MATHEMATICAL model. If you can confirm for me that this is broken we can add that as a new bug to be fixed.

pseudocubic (commented):
Thank you @John-Boik for the contribution, we appreciate it. Overall this PR looks straightforward; I had one question for you, but otherwise it looks good. 

Although there were not tests before, for any new feature we ask that tests be added going forward. Because multinetwork cases are lacking in this repo, I recommend taking a look at the modified [IEEE-13 case inside of PowerModelsONM.jl](https://github.com/lanl-ansi/PowerModelsONM.jl/blob/main/test/data/ieee13_feeder.dss) as a starting point.

src/core/variable.jl
John-Boik: @pseudocubic I removed the check. The issue with the missing `tm_lb` and `tm_ub` occurred because I had used the transformation `remove_all_bounds!` with the engineering model. Some of the examples in in the examples folder used this transformation, and I was following suit. Although I removed the check, some other user might do the same transform for the same reason and then receive the same error. Perhaps the error message should be extended by adding something like "This error can occur if the transformation `remove_all_bounds!` is used." 

I have copied the file `ieee13_feeder.dss` unchanged from PowerModelsONM and placed it in /test/data/opendss. I added a test for the new function `solve_mc_opf_oltc` in the test/multinetwork.jl file. The testset label is also `solve_mc_opf_oltc`. I'm not sure if you want it in the multinetwork test group, or the opf test group. Also, the test values are simply the values that I obtained when running the function. I don't know if they make physical sense or not.

My pull request should pass all tests now.

pseudocubic (approved):
With the unit test added (with the multinetwork tests was the right place for this), it looks good to me.

I see one test failing, but it is unrelated to your additions, so I will accept and merge this.

Also, thanks for looking into the issue with `tm_ub` and `tm_lb`. I will make a note to look at changing the error message.

Thanks for the contribution!

PR #461, opened 2024-07-09
FIX: blanced to balanced

John-Boik:
## Body
This fixes a minor typo in test_cases.jl and test.pf.jl, where "blanced" should be "balanced". The typo has no negative effect, but is fixed here.


pseudocubic (approved):

pseudocubic:
Thank you for the contribution!

